Issue #86, opened 2020-05-30
Add IDE support


PeterJCLaw:
It would be great if we could support competitors continuing to use the IDE for their development if they want to.

For this, we need:
- [x] a script which can build a suitable archive on exporting from the IDE, which will need access to the content of this repo for inclusion (#87)
- [ ] to configure the IDE (either in the IDE itself or our puppet config) to use that script
- [ ] to configure the IDE to use this API as its reference (just using the code from this repo _might_ work, though that should be tested)
- [ ] to configure the IDE for Python 3 code being developed

For the archive script, it needs to be executable with three positional command line arguments and live inside a git repo. These are the [arguments it will be passed](https://github.com/srobo/srobo-ide/blob/master/include/checkout-helper.php#L132):
- the TLA
- the directory containing the user code
- the destination archive to create (will be a `.zip`)

The user code directory will be a folder called `user` within a temporary directory that the script can use as it sees fit. All content of the folder (which may include "hidden", i.e: dot prefixed, files) should be included in the archive.

For context, though this should not be relied upon: the script will be checked out also within the temporary directory, as a sibling of the `user` directory:
```
/tmp/tmp-ide-checkout  # will actually be a randomised `mktmp` style name
├── ilbRobot
│   ├── ...
│   └── .git
└── user
    ├── robot.py
    ├── ...
    └── .user-rev
```

The current directory when the script is run is not defined.

PR #87, opened 2020-05-31
Add a script for use by the IDE

PeterJCLaw:
Contributes to #86.

I've tested locally that this works with an IDE instance.

### DIFF
--- a/script/create-ide-archive
+++ b/script/create-ide-archive
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+"""
+A script for use by the SR IDE to create an archive containing the Simulator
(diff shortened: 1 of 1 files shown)


antoinepetty:
I'm not sure how to review this

trickeydan:
> I'm not sure how to review this

Nor am I. We probably need somebody familiar with the IDE, so possibly @prophile ?

PeterJCLaw:
You could check that it works with a local IDE instance if you like?
The IDE setup instructions in its README should cover you on that side, aside from that all that should be needed is the configuration settings included in my previous comment.

PR #91, opened 2020-06-01
Check for version 3.7 of Python

antoinepetty:
Resolves #89

PR #92, opened 2020-06-01
Python 3.7 CI

RealOrangeOne:
Part of #91


controllers/sr_controller/sr/robot/_version_check.py
```diff
@@ -2,7 +2,7 @@
 
 # Provide a nicer check that the user has the right version of Python than
 # them getting a `SyntaxError`
-assert sys.version_info == (3, 7), (
-    "Sorry, you must be using Python version 3.7. "
+assert sys.version_info >= (3, 7), (
```
antoinepetty, line 5: I think this is wrong, Python 3.8 didn't work for me

PR #96, opened 2020-06-01
Private webot objects

RealOrangeOne:
Protect references to the underlying webots objects,

Where possible, also remove the webots objects from where they're not needed, to clean things up and help even more. Unfortunately that does make the diff a little noisy.

There's little use protecting things further, because at that point it's trying to circumvent the simulation, which isn't doesn't follow good sporting conduct.


.gitignore
```diff
@@ -4,3 +4,4 @@
 /.idea
 /.vscode
 /.mypy_cache
+/venv
```
PeterJCLaw, line 7: Cherry pick to `master` unrelated change?

controllers/sr_controller/sr/robot/motor.py
```diff
@@ -28,33 +28,28 @@ def translate(sr_speed_val, sr_motor):
...
-        -sr_motor.max_speed,
-        sr_motor.max_speed,
+        -sr_motor._max_speed,
+        sr_motor._max_speed,
         sr_speed_val,
     )
 
 
-class Motor(object):
+class MotorBoard(object):
```
PeterJCLaw, line 37: Renaming this class seems unrelated? Is this to match the "real" SR API or just for clarity? Either way, maybe extract this change to simplify this PR?

controllers/sr_controller/sr/robot/sensor_devices.py
```diff
@@ -1,13 +1,19 @@
...
 
 
-class SensorBase(object):
+class SensorBase(ABC):
 
+    @abstractmethod
     def __init__(self, webot, sensor_name):
-        self.webot = webot
-        self.sensor_name = sensor_name
+        pass
```
PeterJCLaw, line 12: Is having this `__init__` in an `ABC` actually useful? I've not used ABCs much, but this doesn't feel to me like it's adding any value.

RealOrangeOne:
In theory, there's little benefit to it being `_webot` over `__webot`. The latter however has slightly more protection around teams abusing webots' own internal API. I believe i've been consistent around them all being `__`, which is good.

controllers/sr_controller/sr/robot/sensor_devices.py
trickeydan: Having used `ABC` a lot in `j5`, I'm pretty confident that `@abstractmethod` has zero effect.

PeterJCLaw: That's odd. My understanding was that the decorator helped prevent you not implementing a method.
For example the following code:
``` python
import abc

class B(abc.ABC):
    def foo(self):
        print('foo')

    @abc.abstractmethod
    def bar(self):
        print('bar')


class C(B):
    def bar(self):
        print('C.bar')

class D(B):
    pass


C()
D()
```

errors like this:
```
Traceback (most recent call last):
  File "Documents\bees.py", line 21, in <module>
    D()
TypeError: Can't instantiate abstract class D with abstract methods bar
```

This is a bit orthogonal though -- my original point was more that (an empty) `__init__` specifically in an ABC doesn't add anything because all classes have one _anyway_.

controllers/sr_controller/sr/robot/motor.py
antoinepetty: I think it's for clarity. The original matches the SR implementation

PeterJCLaw:
> I support blocking team access to the webots objects where possible

Why? What's the motivation here? Does Webots' API allow competitors to do things which could be considered unfair or cheating?

Bear in mind that none of this really prevents the competitors from accessing Webots objects. After all they could just import `controller.Robot` themselves and use that directly, without using our API at all.

In the normal competition we don't place any sort of limitation that teams use our API directly, so I'm not sure why we're doing that here.

PeterJCLaw:
I don't think we're likely to merge this before the competition and there's quite a lot of conflicts here. I'm going to close this and suggest we start afresh if we do want to do something here in the future.

PR #97, opened 2020-06-02
Ignore venv dir

RealOrangeOne:
Cherry-picked from #96

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /.idea
 /.vscode
 /.mypy_cache
+/venv


antoinepetty:
> What are your thoughts on just direct pushing this sort of change? (Perhaps only for "maintainers" of the repo)

I think it's still worth double checking even if just a slack message instead of a pull request

PeterJCLaw:
> > What are your thoughts on just direct pushing this sort of change? (Perhaps only for "maintainers" of the repo)
> 
> I think it's still worth double checking even if just a slack message instead of a pull request

What benefit do you see in asking about this sort of thing?
It feels to me like expressing trust in our colleagues is a better way to go. I'd rather that we enforced correctness through actual checks, rather than review, so really I'm suggesting this when the following are true:
- the person pushing the change is a "maintainer" of the repo, this would be something we limit to a small number of people who have shown they know what they're doing (both generally and specifically have a deep knowledge of the repo in question)
- the repo has branch protection on `master` (prevents force-push messing things up)
- the repo is not auto-deployed (though this case would also be interesting to consider)
- the change is trivial and "obviously correct"
- the repo has CI to catch non-trivial things which turn out to be wrong

To me it feels like a policy of this sort could allow for small things to be fixed with much lower overhead, with very little or zero downside. Is there a negative I'm missing?

antoinepetty:
I don't feel super strongly about this and you raise several good points. I'm happy either way

PR #99, opened 2020-06-04
Multi robot

RealOrangeOne:
Fixes https://github.com/srobo/tasks/issues/594

Allow multiple robots to be run in an arena nicely. By default only the robot in corner 0 has a controller, the other robots will remain stationary. Unfortunately, because of this, it's not possible to distinguish between which robot is which through the logs. This might be possible to do by intercepting `stdout` of the child process, but I don't think it's a huge deal. Can look into if needed.

This PR also uses `.proto` files for the robot, removing the issues of dragging around _part_ of a robot :tada:. Had to do that in this PR as it made my life a whole lot easier!

This PR also handes providing zone and mode information to robots.

Zone is detected automatically, and derived by webots magic IDs for the robots. Using the webot controller caused the controller to crash, which was weird. Should test these work on another machine!

The mode is set using a text file in the root of the repo. `robot_mode.txt` should contain the mode, and defaults to `"dev"` when it doesn't exist.

Although the diff says I have, I haven't really modified the robot at all, apart from allowing variables.


antoinepetty:
What's stopping us having a controller for each robot?

antoinepetty:
The SR logo seems to be missing from the robot now

RealOrangeOne:
> What's stopping us having a controller for each robot?

Mostly code reuse. The code which starts up robot code exists in 1 place: `sr-controller.py`. Multiple controllers means duplicating that code. Unless we have it execute something else in proxy, but more layers means it's more fragile.

PeterJCLaw:
Could you elaborate what you mean by:
> Using the webot controller caused the controller to crash, which was weird.

What do you mean by "the webot controller" here? (Sorry if this is a silly question; I'm still not really familiar with the Webots side of this)

trickeydan:
What is the scope for having team TLAs or something distinguishing on each of the 4 robots so that they can be identified at any point in the match? I feel like 4 identical robots may be an issue.

RealOrangeOne:
@trickeydan the intention is to have the flags on the robot be used for that, as they would in a regular competition.

RealOrangeOne:
@PeterJCLaw if I tried to instantiate the webot controller, it killed the application with status 245 (no stacktrace, nor useful error message). This went away when I didn't instantiate the robot.

What I mean here is trying to use the `controller.Robot` class to get information about the current robot, from inside the `sr_controller.py` file.

antoinepetty:
There are still some failing CI tests

protos/Robot_Echo/SRRobot.proto
```diff
@@ -0,0 +1,1543 @@
...
+                name "solid(8)"
+              }
+              DEF LOGO Solid {
+                translation 0 0.18 0.186
+                children [
+                  Shape {
+                    appearance PBRAppearance {
+                      baseColorMap ImageTexture {
+                        url [
+                          "../robots/Robot_Echo/logo.png"
```
antoinepetty, line 981:
```suggestion
                          "../protos/Robot_Echo/logo.png"
```
Fixes the logo issue

controllers/sr_controller/sr_controller.py
```diff
@@ -7,25 +7,57 @@
...
+    return MODE_FILE.read_text().strip()
+
+
+def main():
+    robot_file = get_robot_file()
+
+    if not robot_file.exists():
         print("Robot controller not found, copying example into place.")
-        copyfile(str(EXAMPLE_CONTROLLER_FILE), str(ROBOT_FILE))
+        copyfile(str(EXAMPLE_CONTROLLER_FILE), str(robot_file))
```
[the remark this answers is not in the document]
PeterJCLaw: :+1: to not copying in comp mode, that's a neat solution.

I don't think it's a good idea to exit cleanly for what is essentially an error though. I suggest that we have the controller error loudly in comp mode if it can't find the `robot.py`. I think we should explicitly signal that we expect no robot in a given zone.

I am very much expecting that we'll also remove robots from the simulation if there isn't a team in that corner, which is what we'd do in a real competition.

controllers/sr_controller/sr_controller.py
```diff
@@ -7,25 +7,68 @@
...
+    if robot_mode == "dev" and not robot_file.exists():
         print("Robot controller not found, copying example into place.")
-        copyfile(str(EXAMPLE_CONTROLLER_FILE), str(ROBOT_FILE))
+        copyfile(str(EXAMPLE_CONTROLLER_FILE), str(robot_file))
+    elif robot_mode == "comp" and not robot_file.exists():
+        print("ERROR: No robot controller found for zone ", robot_zone)
+        sys.exit(1)
+    elif not robot_file.exists():
+        print("No robot controller found for zone ", robot_zone)
+        sys.exit()
```
trickeydan, line 55: Shouldn't we be exiting with a non-zero error code here?

controllers/sr_controller/sr_controller.py
```diff
@@ -7,25 +7,65 @@
...
+    robot_mode = get_robot_mode()
+    robot_zone = get_robot_zone()
+
+    if robot_mode == "dev" and not robot_file.exists():
         print("Robot controller not found, copying example into place.")
-        copyfile(str(EXAMPLE_CONTROLLER_FILE), str(ROBOT_FILE))
+        copyfile(str(EXAMPLE_CONTROLLER_FILE), str(robot_file))
+    elif not robot_file.exists():
+        print("ERROR: No robot controller found for zone ", robot_zone)
+        sys.exit(1)
```
PeterJCLaw, line 52: Just a thought -- would this be clearer if we hoisted the `not robot_file.exists()`?
``` python
    if not robot_file.exists():
        if robot_mode == "dev":
            print("Robot controller not found, copying example into place.")
            copyfile(str(EXAMPLE_CONTROLLER_FILE), str(robot_file))
        else:
            print("ERROR: No robot controller found for zone ", robot_zone)
            sys.exit(1)
```

RealOrangeOne: I was staring at this for ages trying to see how to make it look nicer, that's definitely it!

controllers/sr_controller/sr_controller.py
```diff
@@ -7,25 +7,66 @@
...
+    return ROBOT_IDS_TO_CORNERS[os.environ['WEBOTS_ROBOT_ID']]
+
+
+def get_robot_file() -> Path:
+    zone_id = get_robot_zone()
+    robot_file_dir = ROOT.parent / ("zone-" + str(zone_id))
+    if robot_file_dir.is_dir():
+        return robot_file_dir / "robot.py"
+    if get_robot_mode() == "comp":
+        print("WARNING: Using default robot file location in competition mode.")
```
PeterJCLaw, line 32: Can we simplify this logic?

I suggest that in `comp` mode we don't fall back to the default location -- I can see that causing weirdness during the actual competition otherwise.

RealOrangeOne: It currently prints a warning. In theory it should never happen if the simulator is run in an isolated environment. 

The logic here is a little complex, but I think it's necessary to keep usability simple. Not copying the default controller would simplify things a little, but I think we definitely want to be doing that!

PeterJCLaw: Not sure where we got to with this, especially given the discussion at https://github.com/srobo/docs/pull/126#discussion_r436400754?

Anyway, I've implemented that logic as it seemed not to meet with any objections: https://github.com/srobo/competition-simulator/pull/106

controllers/sr_controller/sr_controller.py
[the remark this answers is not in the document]
PeterJCLaw: acf9e5b6e33d6c391b70211c53fd18cdd6070e67 (merges https://github.com/srobo/competition-simulator/pull/106) reworked this substantially.

controllers/sr_controller/sr_controller.py
PeterJCLaw: acf9e5b6e33d6c391b70211c53fd18cdd6070e67 (merges https://github.com/srobo/competition-simulator/pull/106) reworked this substantially.

PeterJCLaw:
Hrm, there also seems to be a change here such that when running the simulator the perspective is only from the robot. Could we revert that before merging please?

PeterJCLaw:
One other thought, though likely not something to solve here, is how we cope with the case where the user has provided `robot.py` files for zones which do not have robots populated in. This obviously isn't going to be an issue in development, but will be possible in the competition.

A simple solution might be each match always starts with a fresh copy of this repo and then only removes robots from there, rather than needing to add robots. Either way, I think we probably need to be scripting all that, which hopefully reduces the chances of an issue.

antoinepetty:
Is this ready for a merge to Master? Seems to have been sitting here for a while with quite a lot of changes. I suggest making further changes separately

PR #102, opened 2020-06-10
Allow webots controller processes to be isolated using Firejail

RealOrangeOne:
The firejail integration code is at https://github.com/cyberbotics/webots/blob/0275bbe15f7c1f7dc34ea6b9c872e019419e4cfa/src/webots/control/WbController.cpp#L238, and is the only real documentation for it.

The hack around `$SR_SIM_JAIL_PATHS` is because webots currently uses the **process** environment for resolving variables in `runtime.ini` rather than the **controller** environment. I'll probably push a fix upstream at some point.

The jail issues could be resolved by implementing jailing outselves, but I didn't want to do that really, and this way keeps the implementation entirely inside webots itself!

### DIFF
--- a/controllers/sr_controller/runtime.ini
+++ b/controllers/sr_controller/runtime.ini
@@ -1,0 +1,3 @@
+[environment variables with relative paths]
+
+FIREJAIL_PATH=$(SR_SIM_JAIL_PATHS)
--- a/script/webots-firejail
+++ b/script/webots-firejail
@@ -6,7 +6,7 @@ export WEBOTS_FIREJAIL_CONTROLLERS=1
 REPO_ROOT=$(dirname $0)/..
 
 # We want the Firejailed process to have access to the siblings of the
-# simulation directory because the  competitors code lives in sibling directories.
+# simulation directory because the competitors' code lives in sibling directories.
 REPO_PARENT=$REPO_ROOT/..
 
 # Used by `runtime.ini` to pass through all relevant files into jail
--- a/script/webots-firejail
+++ b/script/webots-firejail
@@ -3,7 +3,13 @@
 # Run webots controllers in jails
 export WEBOTS_FIREJAIL_CONTROLLERS=1
 
+REPO_ROOT=$(dirname $0)/..
+
+# We want the Firejailed process to have access to the siblings of the
+# simulation directory because the  competitors code lives in sibling directories.
+REPO_PARENT=$REPO_ROOT/..
+
 # Used by `runtime.ini` to pass through all relevant files into jail
-export SR_SIM_JAIL_PATHS=$(realpath -e ${PWD}/$(dirname $0)/../../)
+export SR_SIM_JAIL_PATHS=$(realpath -e $REPO_PARENT)
 
 exec webots $@
--- a/script/webots-firejail
+++ b/script/webots-firejail
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# Run webots controllers in jails
+export WEBOTS_FIREJAIL_CONTROLLERS=1
+
+# Used by `runtime.ini` to pass through all relevant files into jail
+export SR_SIM_JAIL_PATHS=$(realpath -e ${PWD}/$(dirname $0)/../../)
+
+exec webots $@
--- a/script/webots-firejail
+++ b/script/webots-firejail
@@ -12,4 +12,4 @@ REPO_PARENT=$REPO_ROOT/..
 # Used by `runtime.ini` to pass through all relevant files into jail
 export SR_SIM_JAIL_PATHS=$(realpath -e $REPO_PARENT)
 
-exec webots $@
+exec webots "$@"
--- a/README.md
+++ b/README.md
@@ -28,3 +28,9 @@ On first run, the robot will execute an example program. On first run, this will
 │   └── worlds
 └── robot.py
 ```
+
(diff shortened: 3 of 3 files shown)


PeterJCLaw:
@RealOrangeOne did you mention that you'd submitted a PR upstream to fix the issues you're working around here. Do you think that we're likely to get a release of those fixes any time soon? If so, is it worth waiting until that happens, or do you think we should progress this PR as it stands anyway?

RealOrangeOne:
@PeterJCLaw definitely not. The fix I pushed upstream only partially fixes these issues. I've not written a full patch to deal with them, nor have I worked out what the patch looks like.

Definitely don't think it's worth blocking this until then, especially as what we have here works, well, and it's only minor changes (the simplification of the `webots-firejail` script) which would be done

controllers/sr_controller/runtime.ini
```diff
@@ -0,0 +1,3 @@
+[environment variables with relative paths]
```
PeterJCLaw, line 1: This is a section heading in `ini` format I think? Or is this meant to be a comment? (I think comments in `ini` normally start with `;`?)

script/webots-firejail
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Run webots controllers in jails
+export WEBOTS_FIREJAIL_CONTROLLERS=1
+
+# Used by `runtime.ini` to pass through all relevant files into jail
+export SR_SIM_JAIL_PATHS=$(realpath -e ${PWD}/$(dirname $0)/../../)
```
PeterJCLaw, line 7: I'm pretty sure that use of `$PWD` here is redundant (and possibly incorrect?).

Would it be worth having some interim variables so it's a bit clearer what this does?

```suggestion
REPO_ROOT=$(dirname $0)/..

# We want the Firejailed process to have access to the siblings of the
# simulation directory because the  competitors code lives in sibling directories.
REPO_PARENT=$REPO_ROOT/..

# Used by `runtime.ini` to pass through all relevant files into jail
export SR_SIM_JAIL_PATHS=$(realpath -e $REPO_PARENT)
```

script/webots-firejail
```diff
@@ -0,0 +1,9 @@
...
+
+exec webots $@
```
PeterJCLaw, line 9: To ensure proper re-quoting of any quoted arguments, I think this should instead be:
```suggestion
exec webots "$@"
```

controllers/sr_controller/runtime.ini
RealOrangeOne: Nope, this a section heading, just a _really_ weirdly named one!

RealOrangeOne:
Interesting, I had that issue. Yeah it's caused when it tries to whitelist a path which doesn't exist, or is malformed. Does that directory exist for you?

PeterJCLaw:
Yeah, that path exists (contains things like `libController.so`).

RealOrangeOne:
It's something in the filtering of allowed whitelist files in firejail, although I can't work out what exactly it doesn't like as the code is [verbose](https://github.com/netblue30/firejail/blob/master/src/firejail/fs_whitelist.c). I can reproduce it outside webots with `firejail --debug --whitelist=$HOME true
`, although that doesn't really tell me anything about what's erroring.

Absolutely no idea what to do with this one

PeterJCLaw:
You said you'd had this issue -- what did you do to resolve it?

RealOrangeOne:
The issue was that the `FIREJAIL_PATH` was malformed, specifically the one we're crafting here. I've never had it occur with a directory provided by webots itself.

script/webots-firejail
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Run webots controllers in jails
+export WEBOTS_FIREJAIL_CONTROLLERS=1
+
+REPO_ROOT=$(dirname $0)/..
+
+# We want the Firejailed process to have access to the siblings of the
+# simulation directory because the  competitors code lives in sibling directories.
```
PeterJCLaw, line 9: Double space and grammar. What idiot wrote this?
```suggestion
# simulation directory because the competitors' code lives in sibling directories.
```

PeterJCLaw:
It's not something like we need for our path to include the others, rather than replace them is it?

PeterJCLaw:
I wonder if this will behave more nicely now that we're not doing subprocesses?

PeterJCLaw:
> Having just tried this locally, it doesn't seem to work for me (on Ubuntu 18.04.4 LTS).
> 
> Here's the error I get:
> 
> ```
> INFO: sr_controller: Starting controller: python3 -u "sr_controller.py" 
> [sr_controller] Error: invalid whitelist path /usr/local/webots/lib/controller/
> [sr_controller] Error: proc 19525 cannot sync with peer: unexpected EOF
> [sr_controller] Peer 19526 unexpectedly exited with status 1
> WARNING: 'sr_controller' controller exited with status: 1.
> ```

I'm still getting this on Ubuntu 20.04.1 LTS and Webots R2020b-rev1.

PeterJCLaw:
Closing this as it didn't ever seem to work and would likely need updates given the various OS and Webots changes that have happened since. Notably Webots now has support for controllers connecting via TCP connections, which potentially enables running controllers in e.g: Docker (or even on other machines).

PR #106, opened 2020-06-12
Rewrite the robot file search logic for simplicity and ease of use

PeterJCLaw:
This implements the rules I outlined on the docs PR at https://github.com/srobo/docs/pull/126#discussion_r436400754 while also making the code verbose for clarity.

### DIFF
--- a/controllers/sr_controller/sr_controller.py
+++ b/controllers/sr_controller/sr_controller.py
@@ -68,10 +68,8 @@ def get_robot_file(zone_id: int, mode: str) -> Path:
         print("No robot controller found for zone {}".format(zone_id))
 
         # Only in competition mode is it an error for a robot file to be missing.
-        if mode == "comp":
-            exit(1)
-        else:
-            exit()
+        missing_file_is_error = mode == "comp"
+        exit(1 if missing_file_is_error else 0)
 
     # For the non-strict zones (i.e: Zone 0 in development mode) we check in the
     # fallback place. If that doesn't exist we copy an example into it.
--- a/controllers/sr_controller/sr_controller.py
+++ b/controllers/sr_controller/sr_controller.py
@@ -108,6 +108,8 @@ def main():
     robot_zone = get_robot_zone()
     robot_file = get_robot_file(robot_zone, robot_mode)
 
+    print("Using {} for Zone {}".format(robot_file, robot_zone))
+
     env = os.environ.copy()
     # Ensure the python path is properly passed down so the `sr` module can be imported
     env['PYTHONPATH'] = os.pathsep.join(sys.path)
--- a/controllers/sr_controller/sr_controller.py
+++ b/controllers/sr_controller/sr_controller.py
@@ -18,19 +18,83 @@ ROBOT_IDS_TO_CORNERS = {
     "1470": 3,
 }
 
+STRICT_ZONES = {
+    "dev": (1, 2, 3),
+    "comp": (0, 1, 2, 3),
+}
+
 
 def get_robot_zone() -> int:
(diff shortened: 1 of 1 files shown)


controllers/sr_controller/sr_controller.py
```diff
@@ -18,19 +18,83 @@
...
+        if robot_file.exists():
+            return robot_file
+
+        print("No robot controller found for zone {}".format(zone_id))
+
+        # Only in competition mode is it an error for a robot file to be missing.
+        if mode == "comp":
+            exit(1)
+        else:
+            exit()
```
RealOrangeOne, line 74: This could be inlined and made a bit cleaner
```suggestion
        exit(1 if mode == "comp" else 0)
```

controllers/sr_controller/sr_controller.py
```diff
@@ -18,19 +18,83 @@
...
+
+    # For the non-strict zones (i.e: Zone 0 in development mode) we check in the
+    # fallback place. If that doesn't exist we copy an example into it.
+
+    assert zone_id == 0 and mode == "dev", \
+        "Unexpectedly handling fallback logic for zone {} in {} mode".format(
+            zone_id,
+            mode,
```
RealOrangeOne, line 82: If you're asserting both `zone_id` and `mode` are fixed values, you won't really need the `.format` here.

controllers/sr_controller/sr_controller.py
```diff
@@ -18,19 +18,83 @@
...
+    if fallback_robot_file.exists():
+        return fallback_robot_file
+
+    print("No robot controller found for zone {}, copying example to {}.".format(
+        zone_id,
+        fallback_robot_file,
+    ))
+    copyfile(str(EXAMPLE_CONTROLLER_FILE), str(fallback_robot_file))
+
+    return fallback_robot_file
```
RealOrangeOne, line 97: Inverting the condition here makes things a bit easier to follow

```suggestion
    if not fallback_robot_file.exists():
        print("No robot controller found for zone {}, copying example to {}.".format(
            zone_id,
            fallback_robot_file,
        ))
        copyfile(str(EXAMPLE_CONTROLLER_FILE), str(fallback_robot_file))

    return fallback_robot_file
```

controllers/sr_controller/sr_controller.py
PeterJCLaw: I think we do -- the message is only seen when the values are _not_ the expected fixed values. In that case, the error message includes the unexpected (possibly unknown) values and thus acts as a diagnostic prompt.

controllers/sr_controller/sr_controller.py
PeterJCLaw: I'm not convinced that this is clearer -- you've now got one line which is doing two things.

controllers/sr_controller/sr_controller.py
PeterJCLaw: I'm not convinced that this is necessarily clearer and it opens up possibly having more logic after the example file has been copied in place, which I don't think we want.

controllers/sr_controller/sr_controller.py
PeterJCLaw: Perhaps this might be better?
```suggestion
        missing_file_is_error = mode == "comp"
        exit(1 if missing_file_is_error else 0)
```
Though of course it could be simplified to just:
```suggestion
        missing_file_is_error = mode == "comp"
        exit(missing_file_is_error else)
```
by relying on `True` being `1` and `False` being `0`, though that might be a bit much.

controllers/sr_controller/sr_controller.py
RealOrangeOne: Doh!

PR #108, opened 2020-06-12
Added a bounds check to vision

13ros27:
Added a bounds check to fix the bug raised by me at [https://studentrobotics.org/forum/viewtopic.php?f=10&t=93](url)

### DIFF
--- a/controllers/sr_controller/sr/robot/vision/vectors.py
+++ b/controllers/sr_controller/sr/robot/vision/vectors.py
@@ -12,7 +12,6 @@ DEGREES_TOLERANCE = 10
 class Vector:
     """
     An arbitrary length vector of floating point values.
-
     In addition to the usual Python niceties, this supports scalar
     multiplication & division, vector addition and vector multiplication (dot
     product).
@@ -74,7 +73,6 @@ class Vector:
     def __mul__(self, other: 'Vector') -> float:
         """
         Dot product between two vectors of equal length.
-
         Given vectors A and B, ``A · B == ||A|| ||B|| cos(theta)`` where
         theta is the angle between them.
         """
@@ -107,7 +105,6 @@ class Vector:
 def cross_product(vec_a: Vector, vec_b: Vector) -> Vector:
     """
     Cross product of two 3-vectors.
-
     Given vectors A and B, ``A × B == ||A|| ||B|| sin(theta)`` where
     theta is the angle between them.
     """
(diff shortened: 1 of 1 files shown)


controllers/sr_controller/sr/robot/vision/vectors.py
```diff
@@ -156,6 +156,7 @@ def angle_between(vec_a: Vector, vec_b: Vector) -> float:
     dp = dot_product(vec_a, vec_b)
     mod_ab = vec_a.magnitude() * vec_b.magnitude()
     cos_theta = dp / mod_ab
+    cos_theta = min(1, max(cos_theta, -1))  # Ensures it is within the bounds of 1 and -1
```
PeterJCLaw, line 159: While this is the right domain, I'd much rather have an error than allow possible errors through.

Since we're already attempting to convert the result of the dot product into the right domain it could be that there's something deeper amiss here.

PeterJCLaw:
Ah, I've just seen in the forums you'd already posted that you had a case where the value was just outside the proper domain. I don't suppose you managed to also log the vectors which caused that (so we can have that as a test case?)

13ros27:
Unfortunately I didn't think to record the vectors, I now have that print statement in my code with a try except so that if it happens again I will catch it and print the vectors

13ros27:
I have got the vectors, again with a value of -1.0000000000000002 as cos_theta. `vec_a=Vector((-0.1757748978918965, -0.027978574520328252, -0.09121614242503248))
vec_b=Vector((0.8788744894594824, 0.13989287260164124, 0.4560807121251624))`

controllers/sr_controller/sr/robot/vision/vectors.py
13ros27: `vec_a=Vector((-0.1757748978918965, -0.027978574520328252, -0.09121614242503248)) vec_b=Vector((0.8788744894594824, 0.13989287260164124, 0.4560807121251624))`

PeterJCLaw:
@13ros27 thanks for those example vectors. We're looking at putting together another release of this fairly soon, do you know if you're likely to pick this up soon?

Separately, I was also wondering whether it would make sense to apply a rounding to the number, rather than a hard clamp? That should still allow it to smoothly handle these cases where it's only just out of spec, but wouldn't mask any issues where it's massively wrong.
Perhaps something like:
``` python
     # Ignore occasional floating point rounding errors which push this just outside
     # the valid range (-1 <= cos_theta <= 1), while still erroring for other cases.
     cos_theta = round(cos_theta, 8)
```

PR #110, opened 2020-06-12
Add simulation aware time and sleep methods

### LINKED ISSUE #101: time.sleep uses real time instead of simulation time
time.sleep uses real time instead of simulation time. These are not the same, and as the simulator is being used, we should be using simulation time.

Quoted from https://studentrobotics.org/forum/viewtopic.php?f=10&t=89&p=340#p340

PeterJCLaw:
For reasons of practicality, these are part of the `Robot` object.

Resolves #101.

Built on #109 to placate mypy.


controllers/sr_controller/controller.pyi
```diff
@@ -0,0 +1,7 @@
+# Type stubs for the Webots controller file
```
RealOrangeOne, line 1: Any way this file can be merged with https://github.com/srobo/competition-simulator/blob/master/script/testing/webots-fakes/controller.py?

PeterJCLaw: Ah, possibly.

PeterJCLaw: b0a5b1d19ba9f5164de024c8ca1ad088bc7140ed (see https://github.com/srobo/competition-simulator/pull/109)

PR #113, opened 2020-06-12
Fix boolean logic around continuing step

RealOrangeOne:
Fix issue introduced in #110 

We want to continue when the application *isn't* terminating.

> If the synchronization field is TRUE, this function always returns 0 (or -1 to indicate termination).
> https://www.cyberbotics.com/doc/reference/robot#wb_robot_step

### DIFF
--- a/controllers/sr_controller/sr/robot/robot.py
+++ b/controllers/sr_controller/sr/robot/robot.py
@@ -61,7 +61,7 @@ class Robot(object):
         """
         Run a webots step of the given duration in milliseconds.
 
-        Returns whether or not Webots is about to terminate the simulation.
+        Returns whether or not the simulation should continue (ie is webots terminating).
         """
 
         with self._step_lock:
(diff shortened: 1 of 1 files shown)


controllers/sr_controller/sr/robot/robot.py
```diff
@@ -61,7 +61,7 @@ def webots_step_and_should_continue(self, duration_ms: int) -> bool:
         """
         Run a webots step of the given duration in milliseconds.
 
-        Returns whether or not Webots is about to terminate the simulation.
+        Returns whether or not the simulation should continue (ie is webots terminating).
```
PeterJCLaw, line 64:
```suggestion
        Returns whether or not the simulation should continue (based on
        Webots telling us whether or not the simulation is about to end).
```

PeterJCLaw: Or even just drop the explanation? The inline comment somewhat covers that for anyone who cares.

Issue #116, opened 2020-06-12
Tokens fly around too easily


PeterJCLaw:
As reported in a [couple](https://studentrobotics.org/forum/viewtopic.php?f=9&t=87#p337) of [places](https://studentrobotics.org/forum/viewtopic.php?f=13&t=97&p=371#p371) in the forums, including a [youtube video](https://www.youtube.com/watch?v=gZ1hfbmmyX4).

This potentially means that the tokens will become somewhat challenging just to move around, as well as making it possible for them to get thrown out of the arena.

RealOrangeOne:
This would likely be fixed by some modified version of https://github.com/srobo/competition-simulator/pull/104

PhilR54:
Token mass is currently set far too low at 1g - will implement a fix alongside #100, making it easier to adjust the values in the future.

Issue #100, opened 2020-06-05
Use proto files for tokens


antoinepetty:
Based on the success of using proto files for the robots in #99

Issue #90, opened 2020-06-01
Fix physics engine warnings


antoinepetty:
Currently some errors appear in the console when loading the simulation regarding parts of the robot:
`Webots has detected that this solid is light and oblong according to its inertia matrix. This belongs in the physics edge cases, and can imply weird physical results. Increasing the weight of the object or reducing its eccentricity are recommended.`

Issue #88, opened 2020-06-01
Make material paths relative


antoinepetty:
![image](https://user-images.githubusercontent.com/36314396/83401469-bca16000-a3fc-11ea-93b1-641ed6b9ae68.png)
As highlighted by a team on the forum, the paths to certain textures are Windows-specific. They should work on any platform

TomWheal:
This is also present on Windows, when installed for a single user not an entire machine. 
This could be what students would do on school computers as it doesn't require admin rights.

PeterJCLaw:
Given that it outputs:
> A resource file can be defined relatively to the worlds directory of the current project, relatively to the worlds directory of the default project, relatively to its protos directory (if defined in a PROTO), or absolutely.

It's not obvious to me that any of those options are really what we want. From the looks of it we're trying to use textures from the Webots default projects.

Workarounds therefore include:
- duplicate those textures into our repo or distribution (has licensing questions)
- come up with our own textures
- drop the textures
- something nasty in our controller code which copies them from Webots at runtime

I've not looked into ways to re-use the textures from the default project, though I'm guessing from that error that it's not really a supported case (which seems a shame).

RealOrangeOne:
> relatively to the worlds directory of the default project

Looking at the URL in the image, this is actually what we want. In theory just `./textures/metal.jpg` should work.

I'll try and give this a test tomorrow, but if someone does before even better!

antoinepetty:
I asked on the discord and it seems that you can't create relative paths to the default projects. We'll have to decide to go with one of Peter's options
https://github.com/cyberbotics/webots/issues/1638

antoinepetty:
> > relatively to the worlds directory of the default project
> 
> Looking at the URL in the image, this is actually what we want. In theory just `./textures/metal.jpg` should work.
> 
> I'll try and give this a test tomorrow, but if someone does before even better!

Forgot to mention that I did try this yesterday and it was unsuccessful

PhilR54:
Is this still an issue? I can just remove the metal effect from those parts of the robot - it'll just appear with a matt finish which isn't a whole lot different.

antoinepetty:
Should have been auto closed

PR #117, opened 2020-06-13
Fix robot features

PhilR54:
Aims to address issues with tokens (#116) by adjusting token mass - tokens are also converted to proto format (#100) such that the mass can be adjusted more easily in the future. 

Robot is also tweaked - bounding boxes of bump sensors adjusted to remove physics warnings (#90) and metal parts removed  (#88). The textures for wheels have been made relative and the robot's density has also been increased. 

We should be okay to redistribute the webot's textures - license agreement: https://cyberbotics.com/webots_assets_license


PeterJCLaw:
@PhilR54 thanks for putting this together. One small thing -- please could we keep this to a single topic? I think it's reasonable for the focus here to be on the refactoring around the vervarious warnings, but I'd rather the density fixes (which are actual functionality changes) were a separate PR. Perhaps just remove the last commit (https://github.com/srobo/competition-simulator/pull/117/commits/249807457346c0598b69bebf554b36e6fa97a82e) from this PR so that this just about the warnings (and the refactoring to enable their tidying)?

worlds/Arena.wbt
```diff
@@ -2,8 +2,9 @@
 WorldInfo {
 }
 Viewpoint {
-  orientation 0.33246481952953133 -0.9314016676496985 -0.14818258087426103 5.456433700612629
-  position 6.9972994812935 3.16953523821157 7.1575439818618
+  orientation -0.20519757298593996 -0.9759022170490246 -0.07422141739065735 0.8170417753010643
+  position -8.955236326402577 3.7554386590715487 8.682878887529787
+  follow "S35"
```
PeterJCLaw, line 7: Is this change relevant or can we drop it?

PhilR54: Will be dropped -  left over view point changes when replacing tokens

protos/Robot_Echo/textures/License.md
```diff
@@ -0,0 +1,27 @@
+
+# Textures
+These textures are the property of Cyberbotics:
```
PeterJCLaw, line 3: Would it be possible to either enumerate exactly which files this refers to and/or use a specific directory name for these?
Just in case we add any of our own textures here I think it's worth keeping it clear which are Cyberbotics' and which are not.

PhilR54: I've moved the cyberbotics textures into a new sub-directory, within textures, and referenced the texture files within the license file.  Should be super clear if we add our own now

PhilR54:
> image
> The caster wheel is sinking into the floor a lot of the time for me

With the density change reverted this should now be back to how it was before - it looks like the contact surface was never ideal but I suggest we change the part properties when the density is corrected.

PR #119, opened 2020-06-13
Use symlinked controllers to provide a distinct prefix in the console

PeterJCLaw:
The idea here is that each robot has its own controller, but in the repo they share the same source code.

In theory at least, symlinks are cross platform (ish) having been [part of Windows 10](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/) for a number of years now. Obviously it's possible that some users are using older versions of Windows, though hopefully not many.

The main potential issue I see here is whether the symlinks survive being included in the archive which GitHub creates for us. This probably just needs someone to test unzipping the archive of this branch on Windows 10.

I have checked that this works for a robot.py in the default location (for zone 0) and in the zone-1 location.

### DIFF
--- a/controllers/sr_controller_0/sr_controller_0.py
+++ b/controllers/sr_controller_0/sr_controller_0.py
@@ -1,0 +1,1 @@
+../sr_controller/sr_controller.py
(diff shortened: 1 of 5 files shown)


RealOrangeOne:
The ZIP file GitHub generate from this branch doens't work on windows. It places text files containing the relative path rather than symlinks.

We can solve this by building a zip file ourselves which will replaces the symlinks with actual files, but I don't know whether the symlinks work on windows on a plain checkout

PR #120, opened 2020-06-13
A supervisor robot to run competition matches

### LINKED ISSUE #111: Robot "power" not disabled when game stops
Currently the game over logic just causes the controller to exit, but fails to power off the robot. This means that it will keep doing whatever it was last doing at the end of the game.

While we _can_ pause the simulator, it is likely that we won't be able to pause it at perfectly the right time. It therefore seems likely that we'll end up with cases where robots end up doing stuff after the end of a match.

### LINKED ISSUE #112: Game over logic uses real time, not simulation time
This is essentially a variant of #101, though affecting the competition game-over counter.

We should probably make that game-over logic use simulation time rather than real time.

PeterJCLaw:
Very much WIP, but sharing to give an idea of where this is going.

This resolves #111 as it pauses the whole simulation at the end of the match.
This resolves #112 as it pauses the whole simulation at the end of the match.


controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -0,0 +1,90 @@
...
+# Webots specific library
+from controller import Supervisor  # isort:skip
+
+REPO_ROOT = Path(__file__).resolve().parent.parent.parent
+
+sys.path.append(str(REPO_ROOT / 'controllers/sr_controller'))
+
+import sr_controller  # noqa:E402 # isort:skip
+
+TIME_STEP = 32
```
RealOrangeOne, line 15: I'd be interested in trying to reuse the value in `sr.robot.settings` here if possible.

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -0,0 +1,90 @@
...
+def main() -> None:
+    quit_if_development_mode()
+
+    supervisor = Supervisor()
+
+    prepare(supervisor)
+
+    remove_unused_robots(supervisor)
+
+    # TODO: where should these actually go? Do we configure an understanding of
```
RealOrangeOne, line 83: Saving these based on their start time feels like an accurate way of handling this.

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -0,0 +1,90 @@
...
+    print("Match start")
+    print("===========")
+
+    supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_REAL_TIME)
+
+    duration_ms = TIME_STEP * int(1000 * GAME_DURATION_SECONDS // TIME_STEP)
+    supervisor.step(duration_ms)
+
+    print("==================")
+    print("Game over, pausing")
```
RealOrangeOne, line 68: Given this, should we remove the existing timeout code which exists in the robot process?

PeterJCLaw: Yeah, I think that's the way to go here.

controllers/competition_supervisor/competition_supervisor.py
PeterJCLaw: 2338ae7

controllers/competition_supervisor/competition_supervisor.py
PeterJCLaw: b9d0fbc

controllers/competition_supervisor/competition_supervisor.py
PeterJCLaw: The amount of import path munging needed to achieve that is quite high. 232481c4ceb59888eb2f44f01925aebdb917877a cross links the two instead.

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -0,0 +1,99 @@
...
+def main() -> None:
+    quit_if_development_mode()
+
+    supervisor = Supervisor()
+
+    prepare(supervisor)
+
+    remove_unused_robots(supervisor)
+
+    with record_animation(supervisor, REPO_ROOT / 'recordings' / recording_path()):
```
RealOrangeOne, line 94: Does webots complain if these directories don't exist when it tries to write to them?

PeterJCLaw: It would, so we create them ahead of time. I have however just spotted that we would then error if they already existed, which I've fixed in 23d64d2.

PR #122, opened 2020-06-13
Connect up all robots' controllers

PeterJCLaw:
We have a sensible behaviour in development if there isn't any robot code for these robots and the lack of this is starting to complicate development of the running of the actual competition.


PeterJCLaw:
This is going to conflict with https://github.com/srobo/competition-simulator/pull/119, but we can deal with that later.

PR #126, opened 2020-06-13
Recolour stage

### LINKED ISSUE #103: Remove texture from side of raised platform
Would be good to make the sides a blank colour
image

RealOrangeOne:
Change the colour so the sides don't look weird.

Fixes #103 

image

This is apparently bronze (hah, get it?), although lighting in the screenshot does make it look a bit weird!


antoinepetty:
I was hoping to keep the square outlines for tokens so you can see where they started after robots move them

RealOrangeOne:
Yeah that's sorta nice, but the main reason we do it for resetting, which we don't need here. 

This also removes the need to ensure the tokens are exactly over the drawings, which does look slightly nicer. Although the same could be said for the tokens on the floor.

PeterJCLaw:
What does it look like from the angle we're planning to use as the primary for the livestream?

RealOrangeOne:
@PeterJCLaw screenshot in description

PeterJCLaw:
Do we have the option to move the light source? (I'm assuming we want that viewpoint because that has the arena corners in the "right" places)

antoinepetty:
Given we've decided this is our main viewpoint, I think we have a couple of options. We don't seem to be able to modify the angle of this specific skybox light.

1. Change the skybox light to another one of the preset values. Here's "mountains" - too dark?
image
2. Rotate the entire arena 90 degrees for a more favourable lighting position

antoinepetty:
For reference, here's our current preset: "noon_park_empty"
image

PeterJCLaw:
The "mountains" skybox feels too dark to me, though its hard to know how much of that is relative to the other.
Would it be worth merging `master` here to pick up the floor changes which have since happened?

antoinepetty:
noon_park_empty
image

mountains
image

I can't work out why the new coloured PNG doesn't show shadows but the old one does

antoinepetty:
Also on this branch: My robot IDs no longer match
image

PeterJCLaw:
I wonder if the ids are just internal increments of a counter and the change in texture here removed an element from the list?

JennyLynnFletcher:
I think in terms of looks having the center pedestal being grey looked better than having the bronze/brown

antoinepetty:
> I think in terms of looks having the center pedestal being grey looked better than having the bronze/brown

Here's how it looks with the wall grey
image

PeterJCLaw:
Closing as this is no longer relevant (we're going to have a new game this year).

PR #129, opened 2020-06-14
Fix corner IDs

antoinepetty:
Changes from https://github.com/srobo/competition-simulator/pull/117 stopped the controllers from recognising the corners


PeterJCLaw:
Hrm. Is there another identifier we should be using here? Ideally one which isn't going to change.

PR #132, opened 2020-06-14
Enable control for LEDs

antoinepetty:
Combines branches `/origin/feature/LEDs` and `add-leds`


controllers/sr_controller/sr/robot/ruggeduino_devices.py
```diff
@@ -46,3 +46,17 @@ def __init__(self, webot, sensor_name):
...
+
+class Led(object):
+
+    def __init__(self, webot, device_name):
+        self.webot = webot
+        self.device_name = device_name
+        self.webot_sensor = self.webot.getLED(self.device_name)
+
+    def read_value(self):
+        return self.webot_sensor.get() > 0
```
PeterJCLaw, line 59: What does it mean to read a value from an LED?

controllers/sr_controller/sr/robot/ruggeduino.py
```diff
@@ -17,36 +17,63 @@ def init_ruggeduino_array(webot):
...
-    return [Ruggeduino(webot, analogue_array, digital_array)]
+    return [Ruggeduino(webot, analogue_input_array, digital_input_array, digital_output_array)]
 
 
 class Ruggeduino(object):
 
-    DIGITAL_PIN_OFFSET = 2  # Exclude pins 0 and 1 as they are used for USB serial comms
+    DIGITAL_READ_PIN_START = 2  # Exclude pins 0 and 1 as they are used for USB serial comms
+    DIGITAL_WRITE_PIN_START = 7   # Offset output pins by input pins
+    DIGITAL_END_PIN = 12
```
PeterJCLaw, line 42: This is a reasonable approach, though it might be somewhat more pythonic (and require less effort keeping things in step) to just let invalid requests fail.

It's still possible to provide a useful error message if we want to do that:
``` python
try:
    element = self._list[index - self.OFFSET]
except KeyError:
    raise ValueError("{!r} is not a valid pin (only pins {} - {})".format(
        index,
        self.OFFSET,
        self.OFFSET + len(self._list),
    )) from None

element.whatever()
```
This avoids needing to know exactly what the length of the list is (and thus removes the possibility that it could get out of step) while still providing a clear error message.

controllers/sr_controller/sr/robot/ruggeduino_devices.py
PhilR54: Arduino allows you to read back the output pin, for example quick pin state toggling:
> digitalWrite(pin, !digitalRead(pin))

I'm not familiar with Ruggeduinos firmware, but it's a cool trick.

controllers/sr_controller/sr/robot/ruggeduino.py
```diff
@@ -17,36 +17,63 @@ def init_ruggeduino_array(webot):
...
+        self.digital_input_array = digital_input_array
+        self.digital_output_array = digital_output_array
 
     def digital_read(self, pin):
         "Read an digital input"
-        if pin < 2 or pin > 13:
-            raise ValueError("Only pins 2 - 13 are available on the Ruggeduino")
-        return self.digital_array[pin - Ruggeduino.DIGITAL_PIN_OFFSET].read_value()
+        min_pin = Ruggeduino.DIGITAL_READ_PIN_START
+        max_pin = Ruggeduino.DIGITAL_WRITE_PIN_START - 1
```
PhilR54, line 53: If we're allowing users to read the LED state back the digital read range needs to be changed:

```suggestion
        max_pin = Ruggeduino.DIGITAL_END_PIN 
```

antoinepetty: I got too fed up of trying to deal with these custom messages and arrays so removed the reading LED status functionality. I'm happy for you to revert and fix it instead though

antoinepetty:
I removed the functionality to read the status of an LED for simplicity. I'm happy for someone else to implement that functionality if we still want it

controllers/sr_controller/sr/robot/ruggeduino.py
```diff
@@ -37,9 +37,7 @@ def init_ruggeduino_array(webot):
 
 class Ruggeduino(object):
 
-    DIGITAL_READ_PIN_START = 2  # Exclude pins 0 and 1 as they are used for USB serial comms
-    DIGITAL_WRITE_PIN_START = 7   # Offset output pins by input pins
-    DIGITAL_END_PIN = 12
+    DIGITAL_READ_PIN_START = 2
```
PeterJCLaw, line 40: Any reason to remove the comment here, or should we add that back?
```suggestion
    DIGITAL_READ_PIN_START = 2  # Exclude pins 0 and 1 as they are used for USB serial comms
```

controllers/sr_controller/sr/robot/ruggeduino.py
```diff
@@ -17,36 +17,47 @@ def init_ruggeduino_array(webot):
...
-            raise ValueError("Only pins 2 - 13 are available on the Ruggeduino")
-        return self.digital_array[pin - Ruggeduino.DIGITAL_PIN_OFFSET].read_value()
+        return self.digital_input_array[pin - Ruggeduino.DIGITAL_READ_PIN_START].read_value()
 
     def digital_write(self, pin, level):
-        "Write to an output"
-        raise NotImplementedError("This robot does not support digital_write")
+        "Write a digital output"
+        array_index = pin - Ruggeduino.DIGITAL_READ_PIN_START - len(self.digital_input_array)
+        print(array_index)
```
PeterJCLaw, line 55: Debugging left over?

controllers/sr_controller/sr/robot/ruggeduino.py
RealOrangeOne: This is also no longer just read, so maybe `DIGITAL_PIN_START`

controllers/sr_controller/sr/robot/ruggeduino.py
```diff
@@ -17,36 +17,47 @@ def init_ruggeduino_array(webot):
...
+        self.analogue_input_array = analogue_input_array
+        self.digital_input_array = digital_input_array
+        self.digital_output_array = digital_output_array
 
     def digital_read(self, pin):
         "Read an digital input"
-        if pin < 2 or pin > 13:
-            raise ValueError("Only pins 2 - 13 are available on the Ruggeduino")
-        return self.digital_array[pin - Ruggeduino.DIGITAL_PIN_OFFSET].read_value()
+        return self.digital_input_array[pin - Ruggeduino.DIGITAL_READ_PIN_START].read_value()
```
RealOrangeOne, line 50: This interface is a little weird. On a real board the digital pins are the same list, and you can't have a sensor and LED in the same pin.

It'd be nice to offset these so the LED pins start after the switches.

Alternatively, creating some generic interface for a digital device, and having them implementing read / write. But that is a little more work than a basic offset and check

antoinepetty: > It'd be nice to offset these so the LED pins start after the switches.

This *is* how it currently works. Pins 2-6 are switches and pins 7-12 are LEDs

controllers/sr_controller/sr/robot/ruggeduino.py
```diff
@@ -17,36 +17,46 @@ def init_ruggeduino_array(webot):
...
-        if pin < 2 or pin > 13:
-            raise ValueError("Only pins 2 - 13 are available on the Ruggeduino")
-        return self.digital_array[pin - Ruggeduino.DIGITAL_PIN_OFFSET].read_value()
+        return self.digital_input_array[pin - Ruggeduino.DIGITAL_PIN_START].read_value()
 
     def digital_write(self, pin, level):
-        "Write to an output"
-        raise NotImplementedError("This robot does not support digital_write")
+        "Write a digital output"
+        array_index = pin - Ruggeduino.DIGITAL_PIN_START - len(self.digital_input_array)
```
RealOrangeOne, line 54: If they try to write to a pin which only supports read, this will do weird things. Might be worth adding a check for if `array_index` is negative (ie they're trying to write to an input pin), because otherwise it'll throw a weird looking `KeyError`

PR #135, opened 2020-06-15
Ban accidental usage of 'print'

PeterJCLaw:
Aims to avoid `print` statements from debugging making their way into `master`.

### DIFF
--- a/controllers/sr_controller/sr/robot/game.py
+++ b/controllers/sr_controller/sr/robot/game.py
@@ -10,7 +10,7 @@ def game_over():
     ==========
     Game Over!
     ==========
-    """).strip())
+    """).strip())  # noqa:T001
     _thread.interrupt_main()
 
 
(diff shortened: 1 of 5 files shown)


RealOrangeOne:
The `example_controller.py` contains prints, but they don't seem to be ignored. Is that file being linted (properly)?

PeterJCLaw:
> The `example_controller.py` contains prints, but they don't seem to be ignored. Is that file being linted (properly)?

good spot, it wasn't. I'll fix that.

RealOrangeOne:
My main query was that even though there's a `print` in the `example_controller.py`, tests aren't failing, which makes me think linting isn't being run on it?

PeterJCLaw:
It indeed wasn't. I fixed that separately in https://github.com/srobo/competition-simulator/commit/5dd4364b8c148cf4087c2eab2ca4e51ec585418c.

PR #137, opened 2020-06-15
More type annotations

PeterJCLaw:
This adds some type annotations determined by running [MonkeyType](https://pypi.org/project/MonkeyType/) against a trivial `robot.py` and then applying the results. Not everything is covered here as our stubs are somewhat limited, though hopefully this still provides some value.


controllers/sr_controller/sr/robot/motor.py
```diff
@@ -37,13 +40,13 @@ def translate(sr_speed_val, sr_motor):
 class Motor:
     """A motor"""
 
-    def __init__(self, m0, m1):
+    def __init__(self, m0: Union[Wheel, LinearMotor], m1: Union[Wheel, Gripper]) -> None:
```
RealOrangeOne, line 43: i think we can use the base class for this type rather than building a tightly-coupled union

```suggestion
    def __init__(self, m0: MotorBase, m1: MotorBase) -> None:
```

PeterJCLaw: I thought this too but unfortunately `MotorBase` doesn't have `set_speed`.
An alternative would be to use a `Protocol` here, but that would add a runtime dependency on `typing-extensions`.

RealOrangeOne: eh, leave as-is for now. I'm hoping to try and add some base classes for those motors / sensors properly, at which point we can use _that_ for a type.

trickeydan: I noticed this too. Maybe add an abstract method on MotorBase?

RealOrangeOne: That's my intention, is to `ABC` the lot of them. But that can come later

PR #138, opened 2020-06-15
Move `sr.robot` module outside controllers dir

RealOrangeOne:
Move the `sr.robot` module out of the controllers dir, as it doesn't make much sense in there.


script/linting/lint
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-PY_ROOT=$(dirname $(dirname  $(dirname $0)))/controllers
+PY_ROOT=$(dirname $(dirname  $(dirname $0)))
```
PeterJCLaw, line 3: Perhaps rename this now it's not the _PY_ _ROOT?

(also elsewhere)

script/testing/test
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-cd $(dirname $(dirname  $(dirname $0)))/controllers
+export PYTHONPATH=$(realpath ./stubs):$(realpath ./modules)
 
-export PYTHONPATH=$(realpath ../stub-webots)
-
-python3 -m unittest discover --start-directory sr_controller "$@"
+python3 -m unittest discover --start-directory modules "$@"
```
PeterJCLaw, line 5: Since we're no longer `cd`ing here, `modules` is, I think, an unreliable name.

I suggest sticking with the `cd` approach, even if the paths are updated.

setup.cfg
```diff
@@ -6,7 +6,8 @@ exclude =
     __pycache__,
     build,
     debian,
-    script
+    script,
+    venv
```
PeterJCLaw, line 10: This change seems unrelated? Consider cherry-picking to `master`?

RealOrangeOne: It seemed related, but mostly during a merge fix-up. Without this `flake8` tried to lint the venv, which naturally took ages.

script/testing/test
RealOrangeOne: Not sure I see the difference between `cd`ing into the directory vs pointing `unittest` at it?

script/linting/lint
PeterJCLaw: a548fc685acaefb94c23241940668e5a83f61da1

script/testing/test
PeterJCLaw: 058331824b18530fd728cc7571eba70478755aae

Issue #128, opened 2020-06-13
Clarify what console output is from which robot


PeterJCLaw:
Webots handily prefixed the output from each controller with the name of that controller. However since we're using the same controller for all robots it's hard to tell which robot is outputting what.

It would be handy if there was a way to tell them apart.

PeterJCLaw:
Note: since we raised this issue we have added log files which have the zone id in their name, so this is somewhat mitigated.

PR #140, opened 2020-06-17
Prepare for publication

### LINKED ISSUE #139: Print simulator version number on simulation start up
We intend to release more frequent updates to the simulator. To aid in debugging competitor issues we should output the version number of the simulator code they are running when they start the simulation.

PeterJCLaw:
Suggest reviewing by commit -- this PR contains some sequential changes:
- introduce a script to create archives of this repo which include a version marker (and provides a starting point for further customising the build later)
- have Circle run that when tags are pushed
- have the main controller print the version when run (fixes #139), either from that file or by inspecting the git repo
- have the `Robot` class print the _user's_ code version (from the IDE), if there's a marker file for that

### DIFF
--- a/controllers/sr_controller/sr/robot/robot.py
+++ b/controllers/sr_controller/sr/robot/robot.py
@@ -1,5 +1,6 @@
 import time
 from os import path, environ
+from typing import Optional
 from threading import Lock, Thread
 
 from sr.robot import motor, camera, ruggeduino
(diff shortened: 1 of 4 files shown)


controllers/sr_controller/sr_controller.py
```diff
@@ -101,11 +101,30 @@ def get_robot_mode() -> str:
...
+
+    print("Running simulator version {}".format(version))
+
+
 def main():
     robot_mode = get_robot_mode()
     robot_zone = get_robot_zone()
     robot_file = get_robot_file(robot_zone, robot_mode)
 
+    if robot_zone == 0:
```
RealOrangeOne, line 123: Not sure how I feel about this. It's definitely possible for Zone 0 to be unpopulated (however unlikely), and given the  below `print`, it's not actually reducing any noise.

controllers/sr_controller/sr/robot/robot.py
```diff
@@ -52,6 +52,13 @@ def display_info(self) -> None:
             mode=self.mode,
         ))
 
+        user_version_path = path.join(self.usbkey, '.user-rev')
+        if path.exists(user_version_path):
+            with open(user_version_path) as f:
+                user_version = f.read().strip()
+
+            print("User code: {}".format(user_version))  # noqa:T001
```
RealOrangeOne, line 60: Might be nice to combine this with the above print? Just to reduce noise

controllers/sr_controller/sr_controller.py
PeterJCLaw: This print doesn't convey any more information if it appears four times though, the other one conveys different information per zone.

Bear in mind also that it seems somewhat unlikely that a team will modify their simulation world to remove the zone 0 controller. They're much more likely to move their own code around, which doesn't affect this (as it runs outside user-code space).

Since this is primarily for _our_ convenience I think the noise reduction here is worth it.

controllers/sr_controller/sr/robot/robot.py
PeterJCLaw: I wondered about that, though not for noise reasons -- instead for clarity while #128 is outstanding.

controllers/sr_controller/sr_controller.py
RealOrangeOne: Good point, good enough for me!

controllers/sr_controller/sr/robot/robot.py
PeterJCLaw: 365931f

Issue #114, opened 2020-06-12
Ensure Python processes are cleaned up when the simulation is reset


PeterJCLaw:
A report [on the forum](https://studentrobotics.org/forum/viewtopic.php?f=10&t=95) indicates that sometimes Webots fails to clean up the controller processes.

Possibly related, I've seen times when Webots notes that it's going to force-kill the controller because it didn't exit within one second. The docs imply that when Webots tells you (via return from the `step` method) that the simulation is going to end, that's a courtesy rather than a demand, yet the warning seems to disagree.

I'm not sure if these two are related, however it seems plausible that Webots is actually expecting the controller processes to quit ASAP after it informs them the simulation is going to end (rather than waiting to be killed).

antoinepetty:
This is quite noticeable when running multiple robots. The processes add up quickly

RealOrangeOne:
Looking briefly into how this works:

- Signals aren't sent to the `sr_controller.py` process before hard terminate
- Signals aren't sent to the `robot.py` process before hard terminate

Here's how the hard terminate works: https://github.com/cyberbotics/webots/blob/56bab11b4491c03fd6e9dee8fc64f2ee60b6a701/src/webots/control/WbController.cpp#L157

Also interestingly, the response from `webot.step` doesn't appear to change when the controller is told to stop. This might be because of the non-standard way we're running `step` in a background thread.

PeterJCLaw:
I wonder if we should `exec` rather than `fork` the `robot.py` process?

PeterJCLaw:
An alternative approach than #143 might be to inline execute the competitors' python code within the main controller process.

I'm fairly sure we could do that relatively easily. It would involve:
- setting the working directory to the new value (we already have this to hand anyway)
- reading in their `robot.py`
- calling `exec` on that code (passing empty state)

I would hope that this would avoid the proliferation of child processes and may even help clarify error messages from the child process as it would remove the indirection.

PR #143, opened 2020-06-18
Add handlers to ensure the child process dies

RealOrangeOne:
Attempt to fix https://github.com/srobo/competition-simulator/issues/114

Webots uses SIGTERM on *nix, and SIGKILL on windows. Webots handling code: https://github.com/cyberbotics/webots/blob/56bab11b4491c03fd6e9dee8fc64f2ee60b6a701/src/webots/control/WbController.cpp#L157

### DIFF
--- a/controllers/sr_controller/sr_controller.py
+++ b/controllers/sr_controller/sr_controller.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import signal
 import subprocess
 from shutil import copyfile
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


controllers/sr_controller/sr_controller.py
```diff
@@ -134,14 +135,24 @@ def main():
...
 
-    completed_process = subprocess.run(
+    child_process = subprocess.Popen(
         [sys.executable, "-u", str(robot_file)],
         env=env,
         cwd=str(robot_file.parent),
     )
 
+    def terminate_handler(*args, **kwargs):
+        child_process.terminate()
```
PeterJCLaw, line 145: Should we try to pass on the same signal as we got?

antoinepetty:
It's not working on Windows unfortunately
image

controllers/sr_controller/sr_controller.py
```diff
@@ -134,14 +135,24 @@ def main():
...
+        sys.exit(0)
+
+    # Ensure the child dies
+    signal.signal(signal.SIGKILL, terminate_handler)
+    signal.signal(signal.SIGTERM, terminate_handler)
```
antoinepetty, line 150: This line doesn't seem to work on Windows. However, after removing it the processes still seem to stick around when resetting webots.

trickeydan: Might be worth looking into [PSUtil](https://pypi.org/project/psutil/) for this.

If we don't want to add another dependency, it is BSD licensed, so we can just select the right parts of that library to include in our code.

PeterJCLaw:
Replaced by #150.
Thanks for investigating this approach @RealOrangeOne.

Issue #107, opened 2020-06-12
Math domain error computing visibility of faces


PeterJCLaw:
From https://studentrobotics.org/forum/viewtopic.php?f=10&t=93.

```
Traceback (most recent call last):
  File "robot.py", line 201, in <module>
    markers = R.see()
  File "controllers\sr_controller\sr\robot\camera.py", line 176, in see
    for face in token.visible_faces(is_2d=is_2d):
  File "controllers\sr_controller\sr\robot\vision\tokens.py", line 103, in visible_faces
    return [f for f in faces if f.is_visible_to_global_origin(angle_tolernace)]
  File "controllers\sr_controller\sr\robot\vision\tokens.py", line 103, in <listcomp>
    return [f for f in faces if f.is_visible_to_global_origin(angle_tolernace)]
  File "controllers\sr_controller\sr\robot\vision\tokens.py", line 179, in is_visible_to_global_origin
    angle_to_origin = vectors.angle_between(direction_to_origin, normal)
  File "controllers\sr_controller\sr\robot\vision\vectors.py", line 159, in angle_between
    theta_rads = math.acos(cos_theta)
ValueError: math domain error
```

This potentially implies something rather weird in the vision handling.

PeterJCLaw:
Ok, so (somewhat obviously) the error happens if you pass a value to `math.acos` which is outside the range of a cosine:
``` python
In [6]: math.acos(-1)                                                                                  
Out[6]: 3.141592653589793

In [7]: math.acos(0)                                                                                   
Out[7]: 1.5707963267948966

In [8]: math.acos(1)                                                                                   
Out[8]: 0.0

In [9]: math.acos(1.000001)                                                                            
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-9-dd9a64560939> in <module>
----> 1 math.acos(1.000001)

ValueError: math domain error
```

It's therefore possible that a small float rounding error on the unit vectors could cause this.

I suspect that we might get more information by adding our own check & `ValueError` here to see if anyone hits this again but yielding more useful diagnostic information.

PeterJCLaw:
Note: I suspect that we may also be able to actually remove the angle-check portion of this in a number of its use-cases, since what we really care about is whether or not the vectors are in the same direction. To determine that we can just look at the sign of the value from the dot product, rather than the actual angle. I can't recall whether we have places where we do actually need the angle though.

PR #146, opened 2020-06-20
Prevent floating point rounding errors causing math domain errors

PeterJCLaw:
This introduces a rounding which should have minimal effect on valid inputs, but which will still error for wildly invalid ones.

Fixes https://github.com/srobo/competition-simulator/issues/107.

### DIFF
--- a/modules/sr/robot/vision/vectors.py
+++ b/modules/sr/robot/vision/vectors.py
@@ -156,6 +156,12 @@ def angle_between(vec_a: Vector, vec_b: Vector) -> float:
     dp = dot_product(vec_a, vec_b)
     mod_ab = vec_a.magnitude() * vec_b.magnitude()
     cos_theta = dp / mod_ab
+
+    if abs(cos_theta) > 1:
+        # Round small floating point rounding errors to avoid a math domain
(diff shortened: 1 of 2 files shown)


modules/sr/robot/vision/vectors.py
```diff
@@ -156,6 +156,12 @@ def angle_between(vec_a: Vector, vec_b: Vector) -> float:
     dp = dot_product(vec_a, vec_b)
     mod_ab = vec_a.magnitude() * vec_b.magnitude()
     cos_theta = dp / mod_ab
+
+    if abs(cos_theta) > 1:
+        # Round small floating point rounding errors to avoid a math domain
+        # error from math.acos, without masking genuine errors.
+        cos_theta = round(cos_theta, 15)
```
RealOrangeOne, line 163: Any reason why 15dp? Magic numbers scare me!

PeterJCLaw: Because there are 18 significant figures of precision in a `float` and we should (in theory at least) be looking at a number with magnitude less than 1.

I picked a number which was likely to be sufficiently clear of any rounding errors while still preserving plenty of precision.

PR #148, opened 2020-06-22
Added negative seconds check and error

### LINKED ISSUE #147: sleep doesn't raise error when given negative
If you give the robot.sleep function a negative time interval it doesn't raise an error like the normal time module would (`ValueError: sleep length must be non-negative`) but instead it will just wait for ever. Given how the robot.sleep and robot.time modules are designed to be as close as possible to the time module to use, it should probably raise the same error as the time module does here

13ros27:
Add a check for negative seconds in the sleep function which then raises a value error which should fix #147

### DIFF
--- a/modules/sr/robot/robot.py
+++ b/modules/sr/robot/robot.py
@@ -147,7 +147,6 @@ class Robot:
         """
         Roughly equivalent to `time.sleep` but accounting for simulation time.
         """
-        
         # Checks that secs is positive or zero
         if secs < 0:
             raise ValueError('sleep length must be non-negative')
--- a/modules/sr/robot/robot.py
+++ b/modules/sr/robot/robot.py
@@ -147,6 +147,10 @@ class Robot:
         """
         Roughly equivalent to `time.sleep` but accounting for simulation time.
         """
+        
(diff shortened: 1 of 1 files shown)


antoinepetty:
@PeterJCLaw any idea why the CI isn't coming up here? Is it because the branch is outside the repo?

trickeydan:
> Is it because the branch is outside the repo?

Likely. There's a box you have to tick in CircleCI to build for external repos. It's safe to turn on unless we have any secrets in the CI (e.g for CD), which I don't think we do.

I'd suggest allowing builds of PRs, so that external contributors can have tests run :)

RealOrangeOne:
> Is it because the branch is outside the repo?

It is indeed! Have flicked the switch so it shouldn't happen in future. It won't affect this PR though until new commits are pushed.

modules/sr/robot/robot.py
```diff
@@ -147,6 +147,10 @@ def sleep(self, secs: float) -> None:
         """
         Roughly equivalent to `time.sleep` but accounting for simulation time.
         """
+
```
trickeydan, line 150:
```suggestion

```

PR #150, opened 2020-07-03
Execute the competitor code inline rather than as a subprocess

PeterJCLaw:
This aims to avoid an issue on Windows where the child processes become zombies after the parent is killed forcefully by Webots.

Replaces #143.
Fixes https://github.com/srobo/competition-simulator/issues/114.

### DIFF
--- a/controllers/sr_controller/sr_controller.py
+++ b/controllers/sr_controller/sr_controller.py
@@ -2,7 +2,6 @@ import os
 import sys
 import subprocess
 from shutil import copyfile
-from typing import Dict
 from pathlib import Path
 
 # Root directory of the SR webots simulator (equivalent to the root of the git repo)
@@ -150,9 +149,7 @@ def main():
     # Swith to running the competitor code
     reconfigure_environment(robot_file)
 
-    robot_globals = {}  # type: Dict[str, object]
-
-    exec(robot_file.read_text(), robot_globals, robot_globals)
+    exec(robot_file.read_text(), {})
 
 
 if __name__ == "__main__":
--- a/controllers/sr_controller/sr_controller.py
+++ b/controllers/sr_controller/sr_controller.py
@@ -2,6 +2,7 @@ import os
 import sys
 import subprocess
 from shutil import copyfile
+from typing import Dict
 from pathlib import Path
 
 # Root directory of the SR webots simulator (equivalent to the root of the git repo)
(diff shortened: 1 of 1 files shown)


controllers/sr_controller/sr_controller.py
```diff
@@ -127,21 +142,17 @@ def main():
...
-    sys.exit(completed_process.returncode)
+    # Pass through the various data our library needs
+    os.environ['SR_ROBOT_ZONE'] = str(robot_zone)
+    os.environ['SR_ROBOT_MODE'] = robot_mode
+    os.environ['SR_ROBOT_FILE'] = str(robot_file)
+
+    # Swith to running the competitor code
+    reconfigure_environment(robot_file)
+
+    robot_globals = {}  # type: Dict[str, object]
```
RealOrangeOne, line 153: Any reason to define this upfront rather than passing in 2 constructed dicts to the `exec` call below?

controllers/sr_controller/sr_controller.py
```diff
@@ -115,10 +116,24 @@ def print_simulation_version() -> None:
...
 
 
+def reconfigure_environment(robot_file: Path) -> None:
+    """
+    Reconfigure the interpreter environment for the actual location of the
+    competitor code.
+    """
+
+    # Remove ourselves from the path and insert the competitor code
+    sys.path.pop()
```
RealOrangeOne, line 126: Is it standard for our directory to always be the first? I wonder whether a `.remove` would be safer here?

Alternatively, do we need to remove ourselves?

controllers/sr_controller/sr_controller.py
```diff
@@ -2,6 +2,7 @@
 import sys
 import subprocess
```
RealOrangeOne, line 3: Are we using subprocess anymore in this?

controllers/sr_controller/sr_controller.py
PeterJCLaw: > Is it standard for our directory to always be the first?

Yeah, I believe so. If it wasn't well defined you could end up with some very weird/subtle import bugs on different platforms.

controllers/sr_controller/sr_controller.py
PeterJCLaw: I misread the behaviour of `exec` that we needed to pass the same instance to both `globals` and `locals` in order to get the behaviour we want. Having re-checked that we could pass this inline.

However the other reason I pulled this out is that we can (in theory) use this to override which builtins are available. Maybe not worth having it separate for now.

controllers/sr_controller/sr_controller.py
PeterJCLaw: I wondered this too. We are -- as part of the handling around getting the version of the code being run.

controllers/sr_controller/sr_controller.py
PeterJCLaw: da6a9ed

PeterJCLaw:
I'm happy to merge this, but it would be great if we could validate that this does indeed fix the original issue on Windows.

antoinepetty:
Yep, works on Windows and I no longer see the Python processes which completely ate my CPU. Perfect!

controllers/sr_controller/sr_controller.py
PeterJCLaw: So, it seems that we both missed the actual bug here. This line is the cause of #167 because `pop` changes the _last_ item in the list, not the first.

PR #152, opened 2020-07-03
Attempt to fix the Webots access violations reported in the forums

PeterJCLaw:
As reported [in the forums](https://studentrobotics.org/forum/viewtopic.php?f=10&t=94&start=10), it's possible to get an access violation from within Webots.

I was able to get a similar error locally (by running that team's code) which indicated the source of the error being within the `WbCamreaRecognitionObject_get_model` function, though I suspect that any of the `CameraRecognitionObject` methods could cause this.

From reading through the Webots source for the camera (start around https://github.com/cyberbotics/webots/blob/f704595a9c/src/lib/Controller/api/camera.c#L550) there's a whole bunch of locking apparently going on there, which seems to confirm the theory that we're using objects after they've been cleaned up.

In any case, I was able to reliably reproduce the issue with robot code like this:
``` python
from sr.robot import Robot
R = Robot()
while True:
    R.see()
```

The fix therefore is to pass the step-lock which we already have (currently used for implementing the `Robot.sleep` method safely) into the `Camera` so that the camera can safely do its work within the confines of a single time-step.

### DIFF
--- a/modules/sr/robot/robot.py
+++ b/modules/sr/robot/robot.py
@@ -134,7 +134,8 @@ class Robot:
         self.ruggeduinos = ruggeduino.init_ruggeduino_array(self.webot)
 
     def _init_camera(self) -> None:
-        self.camera = camera.Camera(self.webot)
+        # See comment in Camera.see for why we need to pass the step lock here.
+        self.camera = camera.Camera(self.webot, self._step_lock)
         self.see = self.camera.see
 
     def time(self) -> float:
--- a/modules/sr/robot/camera.py
+++ b/modules/sr/robot/camera.py
@@ -1,5 +1,6 @@
 import re
 import time
+import threading
 from enum import Enum
 from typing import List, Optional, NamedTuple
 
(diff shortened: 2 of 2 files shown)


modules/sr/robot/robot.py
```diff
@@ -134,7 +134,8 @@ def _init_ruggeduinos(self) -> None:
         self.ruggeduinos = ruggeduino.init_ruggeduino_array(self.webot)
 
     def _init_camera(self) -> None:
-        self.camera = camera.Camera(self.webot)
+        # See comment in Camera.see for why we need to pass the step lock here.
+        self.camera = camera.Camera(self.webot, self._step_lock)
```
PeterJCLaw, line 138: I think I'd actually prefer that we didn't pass the lock downwards and instead facaded the `see` method. However since `self.camera` is implicitly part of the API here I think we need to do this this way for now.

I suggest we revisit this at some point and make `self.camera` private (i.e: `self._camera`) and then move the locking back up into an explicit `Robot.see` method.

PeterJCLaw:
I've also asked in the forums if the competitor who was hitting this could try an equivalent fix. I'm going to hold off merging until they respond.

PeterJCLaw:
I've changed my mind on this. I'm going to merge this now as it does seem to fix at least some form of the issue. If we get reports that it's still happening we can revisit.

PeterJCLaw:
The competitor who was experiencing the issue in the forums has confirmed that an equivalent shim fix I asked them to try did indeed work.

For posterity that shim was:
``` python
    def _init_camera(self) -> None:
        self.camera = camera.Camera(self.webot)
        # Remove the line: self.see = self.camera.see

    # Add this method
    def see(self):
        with self._step_lock:
            return self.camera.see()
```

PR #154, opened 2020-07-04
Replace colon with dash for windows html export

### LINKED ISSUE #153: HTML export filename contains invalid characters
Can't have colons in the filename (not on Windows at least)
image

antoinepetty:
Fixes #153

### DIFF
--- a/controllers/competition_supervisor/competition_supervisor.py
+++ b/controllers/competition_supervisor/competition_supervisor.py
@@ -24,6 +24,8 @@ def recording_path(when: Optional[datetime.datetime] = None) -> Path:
 
     date = when.date().isoformat()
     name = '{}.html'.format(when.isoformat())
+    # Windows doesn't like colons in filenames
+    name = name.replace(':', '-')
     return Path(date) / name
 
 


controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -24,6 +24,8 @@ def recording_path(when: Optional[datetime.datetime] = None) -> Path:
 
     date = when.date().isoformat()
     name = '{}.html'.format(when.isoformat())
+    # Windows doesn't like colons in filenames
+    name = name.replace(':', '-')
```
PeterJCLaw, line 28: Hyphens already have a meaning in ISO 8601 formatted datetimes, specifically that they're the `YYYY-MM-DD` part separators. Maybe consider using something else (perhaps dots)?

antoinepetty:
@PeterJCLaw @RealOrangeOne 
would you be happy with a different format such as `%Y%m%dT%H%M%S`?

PeterJCLaw:
:bike: :derelict_house:

I don't think the format matters much as long as it's mostly obvious what it is and doesn't cause errors. Originally I wanted to use match numbers or something like that, so I mostly see dates as a stopgap. Let's ship this PR as is and move on to more important things.

PR #158, opened 2020-07-04
Introduce a separate robot class with manual timestep control

PeterJCLaw:
This allows teams to avoid a variety of timing issues which introduce randomness into the robot behaviours that we don't want. Notably this appears to fix the issues [reported by MAI in the forums](https://www.studentrobotics.org/forum/viewtopic.php?f=9&t=106).

At least the camera needs to have a time step explicitly forced before it can provide useful information, so we handle that in
the Camera itself. This feels reasonable in any cases as we know that a real camera takes a non-zero amount of time to respond. Simple testing suggests that other sensors may not be so affected, though it is hard to know whether that is a natural consequence of them only being used after several timesteps have been done as a matter of course.

----

This is backwards compatible, however when using the new `ManualTimestepRobot` class `time.sleep` alone is no longer suitable for waiting for things to happen -- teams _must_ use `Robot.sleep` instead.
We will need to document both that this new alternative exists and how to use it.

### DIFF
--- a/modules/sr/robot/__init__.py
+++ b/modules/sr/robot/__init__.py
@@ -1,5 +1,5 @@
 import sr.robot._version_check  # noqa
-from sr.robot.robot import Robot
+from sr.robot.robot import Robot, LegacyRobot
 from sr.robot.camera import (
     MarkerType,
     MARKER_ARENA,
@@ -10,6 +10,7 @@ from sr.robot.camera import (
 __all__ = (
     'Robot',
     'MarkerType',
+    'LegacyRobot',
     'MARKER_ARENA',
     'MARKER_TOKEN_GOLD',
     'MARKER_TOKEN_SILVER',
--- a/modules/sr/robot/__init__.py
+++ b/modules/sr/robot/__init__.py
@@ -1,5 +1,5 @@
 import sr.robot._version_check  # noqa
-from sr.robot.robot import Robot, LegacyRobot
+from sr.robot.robot import ManualTimestepRobot, AutomaticTimestepRobot as Robot
 from sr.robot.camera import (
     MarkerType,
     MARKER_ARENA,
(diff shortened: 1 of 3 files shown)


modules/sr/robot/__init__.py
```diff
@@ -12,5 +12,6 @@
     'MarkerType',
     'MARKER_ARENA',
     'MARKER_TOKEN_GOLD',
+    'ManualTimestepRobot',
```
prophile, line 15: Why here in the list?

modules/sr/robot/robot.py
```diff
@@ -158,3 +145,38 @@ def sleep(self, secs: float) -> None:
...
+        # Create a thread which will advance time in the background, so that the
+        # competitors' code can ignore the fact that it is actually running in a
+        # simulation.
+        t = Thread(
+            target=self.webot_run_robot,
+            # Ensure our background thread alone won't keep the controller
+            # process runnnig.
+            daemon=True,
+        )
+        t.start()
```
prophile, line 177: `t` is a bad name for a thread. Could we just call `start` directly on the return value of `Thread`?

modules/sr/robot/robot.py
```diff
@@ -9,8 +9,13 @@
 from controller import Robot as WebotsRobot  # isort:skip
 
 
-class Robot:
-    """Class for initialising and accessing robot hardware"""
+class ManualTimestepRobot:
+    """
+    Class for initialising and accessing robot hardware.
```
prophile, line 14: Minor point: “Class for” contains no information; we can see this is a class by the class keyword above.

modules/sr/robot/__init__.py
PeterJCLaw: They appear to be length sorted.

modules/sr/robot/robot.py
PeterJCLaw: I would agree, however this is moved code which I was aiming to leave unchanged.

modules/sr/robot/robot.py
PeterJCLaw: So I was wondering what the `sr-robot` did any better and it turns out that this is from there.

PeterJCLaw: 1050942

PR #160, opened 2020-07-04
Ensure that all sleep times do actually advance the clock

PeterJCLaw:
Teams don't know what the smallest timestep is, so this ensures that all attempts to sleep do sleep for at least some time. This also avoids issues where the simulation doesn't run because no steps are being taken.

This only really applies when not running the background thread, however as that's the plan anyway (see #158) this seems useful.

### DIFF
--- a/modules/sr/robot/robot.py
+++ b/modules/sr/robot/robot.py
@@ -87,6 +87,11 @@ class Robot:
         Webots telling us whether or not the simulation is about to end).
         """
 
+        if duration_ms <= 0:
+            raise ValueError(
+                "Duration must be greater than zero, not {!r}".format(duration_ms),
+            )
+
         with self._step_lock:
             # We use Webots in synchronous mode (specifically
             # `synchronization` is left at its default value of `TRUE`). In
@@ -151,7 +156,7 @@ class Robot:
             raise ValueError('sleep length must be non-negative')
 
         # Ensure the time delay is a valid step increment
-        n_steps = int((secs * 1000) // self._timestep)
+        n_steps = int((secs * 1000) // self._timestep) or 1
         duration_ms = n_steps * self._timestep
 
         # We're in the main thread here, so we don't really need to do any
--- a/modules/sr/robot/robot.py
+++ b/modules/sr/robot/robot.py
@@ -1,3 +1,4 @@
+import math
 import time
 from os import path, environ
 from typing import Optional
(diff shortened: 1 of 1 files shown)


modules/sr/robot/robot.py
```diff
@@ -151,7 +156,7 @@ def sleep(self, secs: float) -> None:
             raise ValueError('sleep length must be non-negative')
 
         # Ensure the time delay is a valid step increment
-        n_steps = int((secs * 1000) // self._timestep)
+        n_steps = int((secs * 1000) // self._timestep) or 1
```
PeterJCLaw, line 159: The alternative here would be to not do this and have a nicer error at this layer. I've not put a lot of thought into whether that's better here. (Certainly it has the disadvantage that teams may find that code which currently works stops working)

prophile: Should we just error here if you can’t delay for this amount of time, rather than truncating the wait time down to the timestep?

prophile: Alternatively: use an accumulator to minimise the error over time.

modules/sr/robot/robot.py
```diff
@@ -87,6 +87,11 @@ def webots_step_and_should_continue(self, duration_ms: int) -> bool:
         Webots telling us whether or not the simulation is about to end).
         """
 
+        if duration_ms <= 0:
+            raise ValueError(
+                "Duration must be greater than zero, not {!r}".format(duration_ms),
+            )
+
```
prophile, line 94: Perhaps if it’s exactly 0 we exit early?

PeterJCLaw: I'd rather not -- that has impacts on the simulation (that it can get stuck) which this PR is intending to resolve.

modules/sr/robot/robot.py
PeterJCLaw: 00eb22fe6d4bf5838e5680ecf82650a4d31d9b66 resolves this by rounding up rather than down (which is still within the API contract of `time.sleep`).

modules/sr/robot/robot.py
```diff
@@ -150,8 +156,9 @@ def sleep(self, secs: float) -> None:
         if secs < 0:
             raise ValueError('sleep length must be non-negative')
 
-        # Ensure the time delay is a valid step increment
-        n_steps = int((secs * 1000) // self._timestep)
+        # Ensure the time delay is a valid step increment, while also ensuring
```
prophile, line 159: You need to handle the secs=0 case specially I think.

PeterJCLaw: I was actually planning to leave that to error in the underlying method. Do you think it needs handling other than that?

PR #164, opened 2020-07-04
Put recordings in a folder each

### LINKED ISSUE #159: Change recordings to have a folder per recording
Each recording is several files so to make it easier to share them we should group them into a directory per match.

The naming probably wants to stay similar, just add another layer of directory.

PeterJCLaw:
This makes it easier to share each of them (a recording is made of several files).

Fixes #159.

### DIFF
--- a/controllers/competition_supervisor/competition_supervisor.py
+++ b/controllers/competition_supervisor/competition_supervisor.py
@@ -1,7 +1,7 @@
 import sys
 import datetime
 import contextlib
-from typing import Iterator, Optional
+from typing import Iterator
 from pathlib import Path
 
 # Webots specific library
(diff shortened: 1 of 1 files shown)


controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -16,15 +16,13 @@
...
+    now = datetime.datetime.now()
 
-    date = when.date().isoformat()
-    name = '{}.html'.format(when.isoformat())
+    date = now.date().isoformat()
     # Windows doesn't like colons in filenames
-    name = name.replace(':', '-')
-    return Path(date) / name
+    name = str(now.isoformat()).replace(':', '-')
+    return Path(date) / name / '{}.html'.format(name)
```
trickeydan, line 25: Not a criticism, but I'm curious about the usage of `str.format` here over f-strings, as we're working in Python 3.7 for this project.

PeterJCLaw: While the target for the competition is 3.7, we're still supporting 3.5 for development.

PR #165, opened 2020-07-04
Add release instructions

PeterJCLaw:


antoinepetty:
I've not followed all these steps yet. If the process works then great!

Issue #167, opened 2020-07-05
Investigate use of system installed packages in competitor code


PeterJCLaw:
There's a [report in the forums](https://studentrobotics.org/forum/viewtopic.php?f=10&t=103&p=422#p422) that the 0.3.1 release breaks imports of at least `matplotlib`.

PeterJCLaw:
In #150 I changed the sys path which would be available to the competitor code.
I think we may have also done the move of the `sr` module out to `modules` directory as part of `0.3.x` too?

Either way, we should work out why this doesn't work any more as it'll be needed for #163.

PeterJCLaw:
Oh dear. 'Tis a silly mistake.

Turns out that `list.pop()` of course pops from the _end_ of the list. Ooops.

PR #168, opened 2020-07-05
Correct the tidying of sys.path prior to running user code

PeterJCLaw:
`.pop()` pops off the end of the list, not the start. Ooops.

Fixes https://github.com/srobo/competition-simulator/issues/167.

Suggest that we release this as `0.3.2` once merged.

### DIFF
--- a/controllers/sr_controller/sr_controller.py
+++ b/controllers/sr_controller/sr_controller.py
@@ -134,14 +134,9 @@ def reconfigure_environment(robot_file: Path) -> None:
     """
 
     # Remove ourselves from the path and insert the competitor code
-    sys.path[:1] = [
-        # User code first
-        str(robot_file.parent),
-        # Then the SR module
-        str(ROOT / "modules"),
-        # Then everything else which was on the path, except for the
-        # `sr_controller` directory (which was previously first).
-    ]
+    sys.path.pop(0)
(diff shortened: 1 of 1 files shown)


controllers/sr_controller/sr_controller.py
```diff
@@ -134,9 +134,14 @@ def reconfigure_environment(robot_file: Path) -> None:
     """
 
     # Remove ourselves from the path and insert the competitor code
-    sys.path.pop()
-    sys.path.insert(0, str(ROOT / "modules"))
-    sys.path.insert(0, str(robot_file.parent))
+    sys.path[:1] = [
```
RealOrangeOne, line 137: Remind me why we need to remove ourselves from the path?

PeterJCLaw: Only so that they can't import the controller. I suppose it probably doesn't matter much if they can.

RealOrangeOne: I don't think we really need to worry about that. Keeping it as `sys.path.insert` is a lot more readable and approachable.

PeterJCLaw: 3fec9b2 switches back to the simpler spelling.

PR #169, opened 2020-07-06
Added which robot is outputting to print statements

13ros27:
This outputs like this:
`[sr_controller] 0| Print message`
This is potentially annoying if you are only running one robot but I couldn't think of a way to implement a check for that and I also couldn't think of a sensible way to make it toggle-able by the user


13ros27:
This potentially closes #128

controllers/sr_controller/sr_controller.py
```diff
@@ -176,7 +177,10 @@ def __exit__(
 
     def write(self, data):
         self.file.write(data)
-        self.stdout.write(data)
+        if data == '\n':
+            self.stdout.write(data)
+        else:
+            self.stdout.write(f'{self.zone}| {data}')
```
PeterJCLaw, line 183: Unfortunately f-strings aren't supported on Python 3.5, which we do support.

PeterJCLaw, line 183: This is a neat approach, however it won't necessarily work as you're hoping. Specifically it's possible that `write` will be called without a preceding newline having been emitted. This means that code like this:
``` python
print('foo', end='')
print('bar')
```
which we'd want to output "0 | foobar\n" would end up outputting "0| foo0| bar\n".

PR #170, opened 2020-07-06
Rework log file capture to include stderr as well

PeterJCLaw:
This means that any exception traces will also be captured into the log file.

You may wish to review [ignoring whitespace changes](https://github.com/srobo/competition-simulator/pull/170/files?w=1).

### DIFF
--- a/controllers/sr_controller/sr_controller.py
+++ b/controllers/sr_controller/sr_controller.py
@@ -1,11 +1,9 @@
 import os
 import sys
 import datetime
-import contextlib
 import subprocess
-from types import TracebackType
 from shutil import copyfile
-from typing import Type, Optional
+from typing import IO
 from pathlib import Path
 
 # Root directory of the SR webots simulator (equivalent to the root of the git repo)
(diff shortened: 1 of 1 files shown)


controllers/sr_controller/sr_controller.py
```diff
@@ -150,62 +148,61 @@ def log_filename(zone_id: int) -> str:
...
-        self.stack.__enter__()
-
-    def __exit__(
-        self,
-        exctype: Optional[Type[BaseException]],
-        excinst: Optional[BaseException],
-        exctb: Optional[TracebackType],
-    ) -> None:
-        self.flush()
-        self.stack.__exit__(exctype, excinst, exctb)
```
RealOrangeOne, line 175: I notice you're no longer closing the handler on exit, is that intentional?

Obviously during cleanup it doesn't _really_ matter, as the process is dying anyway.

PeterJCLaw: Yes, see the docstring of the new function.

Essentially the issue boils down to what happens in the following code with regards to when the exception trace gets printed to stderr:
``` python
with foo():
    raise SomeError()
```
The steps being:
- `__enter__`
- `raise`
- `__exit__`
- traceback printed

PeterJCLaw: I suspect we might be able to clean up using an at-exit handler, but I don't think it's worth it.
The only thing of consequence which is left around here is the open file handle, which I don't think is an issue.

RealOrangeOne: Ah I see, nice idea!

It's only going to be open slightly longer than normal as the process cleans up and dies, which definitely isn't something we really care about.

PR #171, opened 2020-07-06
Fix Circle archive building

PeterJCLaw:
This uses an interim directory which we remove in the store step to avoid needing to know the exact name of the archive.

As tested by https://app.circleci.com/pipelines/github/srobo/competition-simulator/325/workflows/7d3d861b-16b3-474d-bba9-88351482e6e2/jobs/2556/artifacts

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,11 +128,18 @@ jobs:
       - run:
           name: Build archive
           command: |
+            mkdir archive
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
I had originally envisioned that this step would also do the upload, which was why I didn't have it run every release. Since it doesn't do that, we could just run this all the time, which would avoid the surprise that it was broken.

PR #173, opened 2020-07-07
Remove transparency from png to enable shadows

antoinepetty:
Related to https://github.com/srobo/competition-simulator/pull/126#issuecomment-653738140


antoinepetty:
Example
image

PeterJCLaw:
Please could we also update `textures/makefile` so that recreations of this file match the current output?

PR #180, opened 2020-07-11
Add a script to prepare for a match

PeterJCLaw:
This tells the supervisor what match is about to be played and puts the competitor code in the right places.

### DIFF
--- a/script/prepare-comp-match
+++ b/script/prepare-comp-match
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+"""
+A script to prepare for a competition match.
+"""
+
+import sys
+import shutil
+import argparse
+from pathlib import Path
+from zipfile import ZipFile
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+
+sys.path.append(str(REPO_ROOT / 'controllers/sr_controller'))
+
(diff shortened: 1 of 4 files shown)


script/prepare-comp-match
```diff
@@ -0,0 +1,65 @@
...
+        'archives_dir',
+        help="The directory containing the user's robot code",
+        type=Path,
+    )
+    parser.add_argument('match_num', help="The match number to prepare for.")
+    parser.add_argument(
+        'tla',
+        nargs=NUM_ZONES,
+        help=(
+            "TLA of the team in each zone. Use dash (-) for an empty zone."
```
RealOrangeOne, line 38: Worth noting that TLAs here should be provided in corner order (0, 1, 2, 3)?

I also wonder whether CSV-like would be easier to enter and comprehend than multiple arguments like this, rather verbose.

PeterJCLaw: Yeah, I also wondered about having it consume the path to the compstate or something. I'm not sure to be honest - I was going to for relative simplicity.

Will add a comment about them being in zone order.

PeterJCLaw: 2dff8ee

PR #181, opened 2020-07-11
automatically run video recording

JennyLynnFletcher:
This adds support for automatically capturing video recordings and thus turns match running into a single command.

While not merged here, this PR ends up depending on #205's `wait_start` support in order to simplify the interactions around pausing/running the simulation.


controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -15,7 +15,7 @@
 
 import sr_controller  # noqa:E402 # isort:skip
 
-GAME_DURATION_SECONDS = 150
+GAME_DURATION_SECONDS = 10
```
PeterJCLaw, line 18: debug edit?

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -24,16 +24,22 @@ def recording_path() -> Path:
     date = now.date().isoformat()
     # Windows doesn't like colons in filenames
     name = now.isoformat().replace(':', '-')
-    return Path(date) / name / '{}.html'.format(name)
+    return Path(date) / name / '{}'.format(name)
```
PeterJCLaw, line 27: Perhaps lose the `.format` entirely?

Aside: this is likely to conflict with my changes in #180, but we'll just have to deal with that.

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -24,16 +24,22 @@ def recording_path() -> Path:
...
 
 @contextlib.contextmanager
 def record_animation(supervisor: Supervisor, file_path: Path) -> Iterator[None]:
     file_path.parent.mkdir(parents=True, exist_ok=True)
-    print("Saving animation to {}".format(file_path))
-    supervisor.animationStartRecording(str(file_path))
+    print("Saving animation to {}.html".format(file_path))
+    print("Saving video to {}.mp4".format(file_path))
+    supervisor.animationStartRecording(str(file_path) + '.html')
+    supervisor.movieStartRecording(str(file_path) + '.mp4',
```
PeterJCLaw, line 36: I would suggest that we construct the full names upfront rather than constructing them twice -- that avoids any chance of them getting out of step.

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -24,16 +24,22 @@ def recording_path() -> Path:
...
+                                   width=1920, height=1080,
+                                   quality=100, codec=0, 
+                                   acceleration=1, caption=False)
```
PeterJCLaw, line 39: nit: please could we wrap this to avoid this mixed wrapping style? The local style is like:
```suggestion
    supervisor.movieStartRecording(
        str(file_path) + '.mp4',
        width=1920, 
        height=1080,
        quality=100, 
        codec=0, 
        acceleration=1, 
        caption=False,
    ) 
```

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -118,7 +123,7 @@ def main() -> None:
 
     with record_animation(supervisor, REPO_ROOT / 'recordings' / recording_path()):
         run_match(supervisor)
-
+
```
PeterJCLaw, line 126: whitespace error

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -24,16 +24,22 @@ def recording_path() -> Path:
     date = now.date().isoformat()
     # Windows doesn't like colons in filenames
     name = now.isoformat().replace(':', '-')
-    return Path(date) / name / '{}.html'.format(name)
+    return Path(date) / name / '{}'.format(name)
 
 
 @contextlib.contextmanager
 def record_animation(supervisor: Supervisor, file_path: Path) -> Iterator[None]:
```
PeterJCLaw, line 31: We should probably also rename this.

Or (and from an API/clarity perspective this may be preferable) we could have one context manager for the animation and another for the video. That could be preferable in case we need to turn one of them off for some reason -- it'd just be removing a single context manager to remove one, rather than needing to work out which lines in here should be changed.

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -36,6 +36,21 @@ def record_animation(supervisor: Supervisor, file_path: Path) -> Iterator[None]:
...
+    file_path.parent.mkdir(parents=True, exist_ok=True)
+    print("Saving video to {}".format(file_path))
+    supervisor.movieStartRecording(
+        str(file_path),
+        width=1920,
+        height=1080,
+        quality=100,
+        codec=0,
+        acceleration=1,
+        caption=False)
```
PeterJCLaw, line 50: nit: local style is like
```suggestion
        caption=False,
    )
```

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -119,8 +134,11 @@ def main() -> None:
 
     remove_unused_robots(supervisor)
 
-    with record_animation(supervisor, REPO_ROOT / 'recordings' / recording_path()):
-        run_match(supervisor)
+    recording_path = get_recording_path()
+
+    with record_animation(supervisor, REPO_ROOT / 'recordings' / '{}.html'.format(recording_path)):
+        with record_video(supervisor, REPO_ROOT / 'recordings' / '{}.mp4'.format(recording_path)):
```
PeterJCLaw, line 140: It looks like we've lost a layer of directories here? Bear in mind that the animation outputs several files, so it's advantageous to have a subdirectory per run.

I would also suggest that we pull the construction of the path upwards to reduce the complexity here. I think we want something a bit like this:
```suggestion
    recording_name = get_recording_path()
    recording_path = REPO_ROOT / 'recordings' / recording_name

    with record_animation(supervisor, recording_path / '{}.html'.format(recording_name)):
        with record_video(supervisor, recording_path / '{}.mp4'.format(recording_name)):
```

stubs/controller.py
```diff
@@ -157,6 +157,17 @@ def remove(self) -> None: ...
...
 
+    
+    def movieStartRecording(
+        self,
+        width: int,
+        height: int,
+        quality: int,
+        codec: int,
+        acceleration: int,
+        caption: bool) -> bool: ...
```
PeterJCLaw, line 168: Ah, sorry, I can see you've wrapped this to match the call site.
Since this is a stub file, it's more typical to just allow the longer lines and in this case I think the line is short enough not to worry about it?
```suggestion
    def movieStartRecording(self, width: int, height: int, quality: int, codec: int, acceleration: int, caption: bool) -> bool: ...
```

if that has a linter error I'd be tempted to add a `noqa` comment (with the error id).

stubs/controller.py
```diff
@@ -157,6 +157,17 @@ def remove(self) -> None: ...
     def animationStartRecording(self, file: str) -> bool: ...
     def animationStopRecording(self) -> bool: ...
 
+
```
PeterJCLaw, line 160: nit: I don't think we need two empty lines here
```suggestion
```

stubs/controller.py
PeterJCLaw: Hrm, this line is actually a lot wider than I thought. Wrapping is fine here.

stubs/controller.py
PeterJCLaw: 740b91922bff66afc781d97f811e89a0d7390801

stubs/controller.py
PeterJCLaw: 461b028a4700e09177a239b245bdbc8e90dc44c3

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -90,6 +108,8 @@ def prepare(supervisor: Supervisor) -> None:
     for _, robot in get_robots(supervisor):
         robot.restartController()
 
+    supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_REAL_TIME)
+
```
PeterJCLaw, line 112: This looks like a behaviour change -- is this intentional?

PeterJCLaw:
https://github.com/cyberbotics/webots/blob/b210acad220166fe6be395f8f10ede6614e0ba75/src/webots/gui/WbVideoRecorder.cpp#L144 seems to be the source the message we were seeing.

From my own testing, it seems to be the reset which causes the issue, which isn't ideal given that we kinda need it given the lack of a `wait_start` solution. I do also think there's maybe some issues in Webots here though too.

PeterJCLaw:
I've raised https://github.com/cyberbotics/webots/issues/2386 for the issue of the simulation reset causing capture issues.

controllers/competition_supervisor/competition_supervisor.py
PeterJCLaw: This is related to ensuring that we do actually capture video, relates to https://github.com/cyberbotics/webots/issues/2386.

controllers/competition_supervisor/competition_supervisor.py
PeterJCLaw: cb818d4

controllers/competition_supervisor/competition_supervisor.py
PeterJCLaw: This becomes moot once we have `wait_start` support from #205.

README.md
```diff
@@ -87,19 +87,16 @@ In order to run competition matches you'll need to:
...
+   then pause the simulation at the end.
 
-8. After match complete robots will stop moving, press esc.
-   Wait for video to finish processing and saving before closing webots
+6. After the match completes the robots will stop moving. Wait for the video to
+   complete its rendering (check in the output console) and then quit Webots.
 
-9. Copy logs into `competition-simulator/recordings/yyy-mm-dd/match-<match number>/logs`
-
-10. Close webots and repeat for next match
```
RealOrangeOne, line 105: Is this step not required anymore? How are subsequent matches done?

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -147,8 +171,11 @@ def main() -> None:
 
     remove_unused_robots(supervisor)
 
-    with record_animation(supervisor, REPO_ROOT / 'recordings' / recording_path()):
-        run_match(supervisor)
+    recording_path = REPO_ROOT / 'recordings' / get_recording_path()
```
RealOrangeOne, line 174: Should `get_recording_path` just return the fully-formed path? If not I wonder if there's a better name for it?

README.md
PeterJCLaw: Ah. The repeat is needed, just not the closing. I failed to read the whole thing before removing.

PeterJCLaw: 796c48a clarifies that the procedure is for a single match. I think it's fairly obvious how to generalise it.

controllers/competition_supervisor/competition_supervisor.py
PeterJCLaw: Maybe? Do you have a suggestion for a better name?

I've also been thinking about whether this sort of thing should be more configurable. I think eventually I'd like to get to a point where there's either a configuration file and/or suite of environment variables which we can use to control where things like the recordings end up in order to more completely automated the running of the competition. I'm therefore not especially bothered about these being imperfect for now (i.e: "good enough" is fine).

RealOrangeOne (approved):

PR #183, opened 2020-07-12
Add a supervisor log

PeterJCLaw:
Since the discovery of Webots' `--stdout`, `--stderr` and `--minimise` flags this now feels much less necessary, however I think it's potentially still useful.
~Certainly this _isn't_ a requirement for match automation any more.~

### DIFF
--- a/controllers/competition_supervisor/competition_supervisor.py
+++ b/controllers/competition_supervisor/competition_supervisor.py
@@ -18,6 +18,11 @@ import controller_utils  # isort:skip
 GAME_DURATION_SECONDS = 120
 
 
+def get_log_filename() -> Path:
+    identifier = controller_utils.get_filename_safe_identifier()
+    return controller_utils.ARENA_ROOT / f'supervisor-log-{identifier}.txt'
+
+
 def get_recording_path() -> Path:
     now = datetime.datetime.now()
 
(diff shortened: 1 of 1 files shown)


controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -18,6 +18,13 @@
 GAME_DURATION_SECONDS = 150
 
 
+def log_filename() -> str:
+    # TODO: maybe merge with `recording_path`?
```
PeterJCLaw, line 22: Note: I'm not doing this here because we expect that #181 will change the way that `recording_path` works.

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -120,6 +127,8 @@ def run_match(supervisor: Supervisor) -> None:
 def main() -> None:
     quit_if_development_mode()
 
+    sr_controller.tee_streams(REPO_ROOT / 'recordings' / log_filename())
```
RealOrangeOne, line 130: Is 'recordings' the best directory for this?

PeterJCLaw: Probably not. I'm open to ideas for better places to put it. I was mostly aiming for somewhere which was already somewhere that you might look when running competition mode but wasn't going to get in the way.

PeterJCLaw:
https://github.com/srobo/competition-simulator/pull/256 also includes such a log, so if that merges this will be redundant.

PeterJCLaw:
#256 has merged, this is redundant.

PR #185, opened 2020-07-12
add script to put logs into buckets per team name

Adimote:
This was used to generate the team log files sent to the mentors

### DIFF
--- a/script/bucket-logs-per-team
+++ b/script/bucket-logs-per-team
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+
+import glob
+import os
+import sys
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
I'm removing this from the milestone for a couple of reasons:
- I don't think we need this for Kickstart, and
- I think we would be better served by more generally rethinking how we run matches, rather than necessarily pushing forwards with automating the exact steps we used last year

I've created #210 for a discussion on the second point.

PeterJCLaw:
Closing as this is replaced by #189 and is implemented in #256.

PR #188, opened 2020-07-18
Introduce a maximum frequency for LED changes

PeterJCLaw:


modules/sr/robot/ruggeduino_devices.py
```diff
@@ -43,6 +43,16 @@ class Led:
 
     def __init__(self, webot, device_name):
         self.webot_sensor = webot.getLED(device_name)
+        self._webot = webot
```
RealOrangeOne, line 46: nit: If we're keeping reference to `webot`, might as well use it everywhere

```suggestion
        self._webot = webot
        self.webot_sensor = self._webot.getLED(device_name)
```

prophile:
This is a maximum frequency, is it not?

Adimote:
Tested with and without this change:

Before:
unlimited

After:
limited

modules/sr/robot/ruggeduino_devices.py
PeterJCLaw: I'm not sure what benefit this provides? I think either spelling is fine, so I'd rather err on the side of leaving existing code as-is.

PeterJCLaw:
One thing which does occur to me now is that there are several LEDs on the robot, do we need to ensure that it's only four changes overall per second?

trickeydan:
> One thing which does occur to me now is that there are several LEDs on the robot, do we need to ensure that it's only four changes overall per second?

In the interests of what this PR is for, it should limit over all LEDs inclusively.

RealOrangeOne:
That point also raises the far more difficult implementation: Should this limit be synced between all robots?

PeterJCLaw:
> That point also raises the far more difficult implementation: Should this limit be synced between all robots?

This would be _much_ harder to do. I suggest that we focus on getting the robot-level limit in place and come back to that separately.

modules/sr/robot/ruggeduino_devices.py
```diff
@@ -41,18 +45,19 @@ def read_value(self):
...
-        self._last_change = 0
+        self._limiter = limiter
 
     def write_value(self, value):
-        now = self._webot.getTime()
-        diff = now - self._last_change
-        if diff < 0.6:
-            # Avoid creating strobing effects by allowing LEDs to change at most
-            # once per second.
+        if not self._limiter.can_chage():
```
trickeydan:
```suggestion
        if not self._limiter.can_change():
```

PeterJCLaw: I think https://github.com/srobo/competition-simulator/pull/188/commits/ba3f4cb728dc966332918a02e7a1541f4881c6b4 does this?

modules/sr/robot/output_frequency_limiter.py
```diff
@@ -0,0 +1,23 @@
+from controller import Robot
+
+# Allow up to two changes per second
+MAX_FREQUENCY = 2
```
RealOrangeOne, line 4: Should the frequency be an argument to the limiter?

PeterJCLaw: I thought about that but decided it would be easy enough to add that support when we needed it.

RealOrangeOne (approved):

PR #196, opened 2020-09-28
Adds much inline documentation

### LINKED ISSUE #44: Improve code documentation
The code I've been writing has barely any comments. Would be good to add more for future developers

antoinepetty:
Fixes #44

This isn't complete by any means, but is a useful step towards that.
This adds a bunch of documentation to our internals as a mixture of docstrings type signatures and comments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 /.mypy_cache
 /venv
 /recordings
+.vscode/settings.json
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,3 @@
 /.mypy_cache
 /venv
 /recordings
-.vscode/settings.json
--- a/modules/sr/robot/motor.py
+++ b/modules/sr/robot/motor.py
@@ -75,3 +75,16 @@ class MotorChannel:
             value = -SPEED_MAX
 
         self.sr_motor.set_speed(translate(value, self.sr_motor))
+
+    ''''@property
+    def use_brake(self):
+        "Whether to use the brake when at 0 speed"
+        return self._use_brake
+
+    @use_brake.setter
+    def use_brake(self, value):
+        self._use_brake = value
+
+        if self.power == 0:
+            "Implement the new braking setting"
+            self.power = 0'''
--- a/modules/sr/robot/motor.py
+++ b/modules/sr/robot/motor.py
@@ -76,15 +76,3 @@ class MotorChannel:
 
         self.sr_motor.set_speed(translate(value, self.sr_motor))
 
-    ''''@property
-    def use_brake(self):
-        "Whether to use the brake when at 0 speed"
-        return self._use_brake
-
-    @use_brake.setter
-    def use_brake(self, value):
-        self._use_brake = value
-
-        if self.power == 0:
-            "Implement the new braking setting"
-            self.power = 0'''
--- a/modules/sr/robot/motor.py
+++ b/modules/sr/robot/motor.py
@@ -75,4 +75,3 @@ class MotorChannel:
             value = -SPEED_MAX
 
         self.sr_motor.set_speed(translate(value, self.sr_motor))
-
--- a/modules/sr/robot/motor_devices.py
+++ b/modules/sr/robot/motor_devices.py
@@ -2,6 +2,10 @@ from controller import Robot
 
 
 class MotorBase:
+    """
+        A base class for any type of motor as these are common attributes
+    """
+
     def __init__(self, webot: Robot, motor_name: str) -> None:
         self.motor_name = motor_name
         self.webot_motor = webot.getMotor(motor_name)
@@ -9,6 +13,9 @@ class MotorBase:
 
 
 class Wheel(MotorBase):
+    """
+        A standard rotational motor in Webots
+    """
 
     def __init__(self, webot: Robot, motor_name: str) -> None:
         super().__init__(webot, motor_name)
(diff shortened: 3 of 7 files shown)


modules/sr/robot/ruggeduino_devices.py
```diff
@@ -30,6 +34,10 @@ def read_value(self):
 
 
 class Microswitch:
+    """
+    A standard Webots touch sensor.
+    Reading from this sensor returns a boolean to match the SR API.
```
PeterJCLaw, line 39: Does the SR API have microswitches specifically? I wasn't aware of that.

modules/sr/robot/ruggeduino_devices.py
```diff
@@ -4,6 +4,10 @@
 
 
 class DistanceSensor:
+    """
+    A standard Webots distance sensor. Unfortunately there is a 30cm range limit within Webots.
+    We convert the distance to metres as to match the standard SR API.
```
PeterJCLaw, line 9: Does the SR API have ultrasound sensors built-in? I'm not aware that it does (though j5 and the SourceBots APIs may do).

modules/sr/robot/randomizer.py
```diff
@@ -1,6 +1,11 @@
 import random
 from typing import TypeVar
 
+"""
+    To better reflect reality, we add a little randomness to robot inputs and outputs.
+    This prevents the simulation from providing unrealistic perfection.
+"""
```
PeterJCLaw, line 7: Module docstrings should come before imports.

modules/sr/robot/motor_devices.py
```diff
@@ -2,13 +2,20 @@
 
 
 class MotorBase:
+    """
+        A base class for any type of motor as these are common attributes
```
PeterJCLaw, line 6: Style: I suggest not indenting the content of a docstring more than the quotes are:
```suggestion
    A base class for any type of motor as these are common attributes
```
(also elsewhere)

modules/sr/robot/motor_devices.py
```diff
@@ -36,6 +46,13 @@ def set_speed(self, speed):
 
 
 class Gripper(MotorBase):
+    """
+        Our default robot's gripper has its fingers attached to linear motors.
+        This class allows you to control these as a single entity to open/close.
+
+        You should initialise with the motor_name as the two Webots motor names separated by
+        a pipe character (|)
```
PeterJCLaw, line 54: Hrm, I'd not spotted this. We should perhaps change this to being a 2-tuple rather than requiring string parsing? Not for this PR though.

modules/sr/robot/ruggeduino.py
```diff
@@ -5,6 +5,10 @@
 
 
 def init_ruggeduino_array(webot: Robot) -> 'List[Ruggeduino]':
+    """
+    The names in these arrays correspond to the names given to devices in Webots
+    """
+
```
PeterJCLaw, line 11: Suggest that this be a comment rather than a docstring -- the detail doesn't matter to callers, nor describes how to use this function.

.gitignore
```diff
@@ -9,3 +9,4 @@
 /.mypy_cache
 /venv
 /recordings
+.vscode/settings.json
```
PeterJCLaw, line 12: I don't think this is needed? We already ignore `/.vscode` as a whole folder. Did you have a case where this wasn't being ignored?

PeterJCLaw (commented):
Nice, thanks for putting this together :).
Some small suggestions and questions inline.

modules/sr/robot/ruggeduino_devices.py
trickeydan: > Does the SR API have microswitches specifically?

Definitely doesn't.

antoinepetty: This module is a halfway point between Webots and the SR API. Perhaps the docstring was misleading by mentioning the SR API so I'll delete that part

.gitignore
antoinepetty: I did have a case where the settings file still wanted to be added, but this didn't fix it so there's not much point

modules/sr/robot/motor.py
```diff
@@ -75,16 +75,3 @@ def power(self, value):
             value = -SPEED_MAX
 
         self.sr_motor.set_speed(translate(value, self.sr_motor))
-
```
PeterJCLaw, line 78: Unrelated change? I realise this is dead code, but this doesn't feel like the right PR to be removing it in.

modules/sr/robot/ruggeduino.py
```diff
@@ -5,6 +5,9 @@
 
 
 def init_ruggeduino_array(webot: Robot) -> 'List[Ruggeduino]':
+
```
PeterJCLaw, line 8: I'd probably drop this empty line.

antoinepetty, line 8:
```suggestion
```

modules/sr/robot/utils.py
```diff
@@ -1,2 +1,9 @@
-def map_to_range(old_min, old_max, new_min, new_max, value):
+def map_to_range(
+    old_min: float,
```
RealOrangeOne, line 2: Is there a need to constrain this to operating on `float` over also `int`?

RealOrangeOne (approved):

modules/sr/robot/utils.py
PeterJCLaw: Due to what this is doing the output will always be `float`, so there's nothing to be gained from generics. Since you can always pass an `int` to something expecting a `float` this is simplest to define.

modules/sr/robot/motor_devices.py
PeterJCLaw: #198 fixes this.

PR #198, opened 2020-10-24
Replace string parsing with tuple literal

PeterJCLaw:
This simplifies the code as well as allowing the type signature to specify exactly how many entries should be present.

### DIFF
--- a/modules/sr/robot/motor_devices.py
+++ b/modules/sr/robot/motor_devices.py
@@ -1,3 +1,5 @@
+from typing import Tuple
+
 from controller import Robot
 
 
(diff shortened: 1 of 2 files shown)


modules/sr/robot/motor_devices.py
```diff
@@ -37,12 +39,10 @@ def set_speed(self, speed):
 
 class Gripper(MotorBase):
 
-    def __init__(self, webot: Robot, motor_name: str) -> None:
+    def __init__(self, webot: Robot, motor_names: Tuple[str, str]) -> None:
```
RealOrangeOne, line 42: I wonder if instead, this should just be 2 separate arguments?

PeterJCLaw: Eh. They do en up in a list and are treated as just elements of that list though, so it feels more fitting to accept them like this.

RealOrangeOne (approved):

PR #201, opened 2020-10-25
Promote the ManualTimestepRobot

PeterJCLaw:
This requires competitor code to explicitly use our API wrappers when they want time to advance. In turn this should make robot behaviour more predictable and allows us to simplify how we're running matches (as we no longer need to ensure they run at 1:1 real time).

### DIFF
--- a/modules/sr/robot/robot.py
+++ b/modules/sr/robot/robot.py
@@ -1,8 +1,7 @@
 import math
-import time
 from os import path, environ
 from typing import Optional
-from threading import Lock, Thread
+from threading import Lock
 
 from sr.robot import motor, camera, ruggeduino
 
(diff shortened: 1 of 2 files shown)


modules/sr/robot/robot.py
```diff
@@ -152,38 +151,3 @@ def sleep(self, secs: float) -> None:
...
-            target=self.webot_run_robot,
-            # Ensure our background thread alone won't keep the controller
-            # process runnnig.
-            daemon=True,
-        )
-        t.start()
-        time.sleep(self._timestep / 1000)
-
-    def webot_run_robot(self):
-        while self.webots_step_and_should_continue(self._timestep):
```
RealOrangeOne, line 188: I wonder whether we should remove the `_and_should_continue` functionality?

PeterJCLaw: I can see it potentially being useful for competitors. I have fixed a related comment in 89ea239 though.

RealOrangeOne (approved):

PR #204, opened 2020-10-25
Improve developer UX

PeterJCLaw:
This primarily aims to improve the UX for those developing the simulator itself, though the logging changes will benefit competitors too.

I also expect that the ability to control the `ARENA_ROOT` is likely to also be helpful when running the competitions. For example we could easily now move the `recordings` folder to being within the `ARENA_ROOT` so that each such directory contains everything for a given match.

### DIFF
--- a/controllers/sr_controller/sr_controller.py
+++ b/controllers/sr_controller/sr_controller.py
@@ -11,10 +11,10 @@ REPO_ROOT = Path(__file__).resolve().parent.parent.parent
 EXAMPLE_CONTROLLER_FILE = REPO_ROOT / 'controllers/example_controller/example_controller.py'
 
 # Root directory of the specification of the Arena (and match)
-ARENA_ROOT = REPO_ROOT.parent
+ARENA_ROOT = Path(os.environ.get('ARENA_ROOT', REPO_ROOT.parent))
 
-MODE_FILE = REPO_ROOT / 'robot_mode.txt'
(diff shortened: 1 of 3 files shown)


controllers/sr_controller/sr_controller.py
```diff
@@ -176,10 +179,28 @@ class SimpleTee:
...
+    def __init__(self, *streams: IO[str], prefix: str = '') -> None:
         self.streams = streams
+        self._line_start = True
+        self.prefix = prefix
+
+    def _insert_prefix(self, data: str) -> str:
+        # Append our prefix just after all inner newlines. Don't append to a
+        # trailing newline as we don't know if the next line in the log will be
+        # from this zone.
+        final_newline = data.endswith('\n')
```
RealOrangeOne, line 191: Do we know how this performs on windows, given that generally uses `\r\n` for newlines

PeterJCLaw: I'm assuming you mean this question about the following line (where we do the `replace`) rather than this line? Obviously this line will be completely fine on Windows.
---
So Python uses universal newlines -- any time it reads in or wants to handle a newline you _should_ get `\n` rather than `\r\n`. Obviously this doesn't stop someone putting a literal `\r\n` in their code and printing that and in that case this will likely get weird, however that code is then not portable and would have weird output on Unix anyway.

RealOrangeOne (approved):

PR #205, opened 2020-10-25
Implement wait_start via Webots customData field

### LINKED ISSUE #200: Fully support wait_start
I think we're going to need to have a proper mechanism to have the robots wait for the start of the match if we want to have automatic video collection (#181) due to limitations in webots (https://github.com/cyberbotics/webots/issues/2386).

Since we previously ran the competition on Windows I'm assuming that we need to support doing this on Windows too. Obviously we'd really like that anyway in order to support development in competition mode on Windows.

The way that this works on the real robots is that a FIFO is created using `os.mkfifo`, however that's a Unix-only API so that's out.

The other alternatives I can think of are:
- **use a `socket.AF_UNIX` socket**
  Impractical as this isn't supported yet in Python (https://bugs.python.org/issue33408). It looks like it might land some time during Python 3.9, but it's not here yet.
- **use a Named pipe**
  [pywin32](https://pypi.org/project/pywin32/) seems to be the consensus for using these on Windows, however that means adding a potentially complicated (and undocumented) dependency along with factoring out the backends to be platform specific
- **watch a directory for a file appearing (or disappearing)**
  This is horrible.

Any other ideas?

PeterJCLaw:
This is closer to how we really run matches and removes the need to reset the simulation as part of the setup process.

In turn, this will make it possible to automate video capture of the matches without running into Webots limitations.

For the moment I've left the animation capture as starting at the start of the match time, however it could now start earlier (i.e: including the setup phase too) if we wanted.

Suggest review by commit as I needed to fix some type annotations first.
~Note: I've based this on `improve-dev-ux` (#204) for my ease of development, though there isn't actually a dependency here.~

Fixes #200


controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -104,6 +115,11 @@ def run_match(supervisor: Supervisor) -> None:
     print("Match start")
     print("===========")
 
+    # First signal the robot controllers that they're able to start ...
+    for _, robot in get_robots(supervisor):
+        robot.getField('customData').setSFString('start')
```
RealOrangeOne, line 120: Thoughts on making `"start"` and `"ready"` some form of constant?

PeterJCLaw: Maybe, though I expect that the import shenanigans we'd need to be able to share them would be more trouble than it's worth.

RealOrangeOne: Yeah that's true. There's still some benefit to defining them in a constant place. We can use comments to ensure they're updated consistently in future.

PeterJCLaw: c475411

RealOrangeOne (approved):

PR #208, opened 2020-11-01
Add types everywhere

PeterJCLaw:
This adds type annotations to the remaining few places they were missing and then updates the config to require them everywhere.


modules/sr/robot/ruggeduino.py
```diff
@@ -69,7 +69,7 @@ def analogue_read(self, pin: int) -> float:
         """Read an analogue input"""
         return self.analogue_input_array[pin].read_value()
 
-    def pin_mode(self, pin_no, mode):
+    def pin_mode(self, pin_no: int, mode: object) -> None:
```
trickeydan, line 72: Please use `typing.Any` instead of object. It makes it much easier to remove unknown types as mypy can warn with the right config.

trickeydan (changes requested):

modules/sr/robot/ruggeduino.py
PeterJCLaw: Do you have an example of where that would matter? I've recently started using `object` to signal that a parameter is unused, as distinct from `Any` which I tend to use when I'm passing through a value to a wrapped function. If that creates issues it would be useful to understand them.

RealOrangeOne: I'd have to agree with using `Any`, even if just for the sake of not confusing people when they read it.

Given we're already setting up pins for them, and the API is already not compatible, I wonder if we even need this method stub at all. Can't have semantics issues if you don't have any code!

PeterJCLaw: Would be good to reach a consensus on this before it diverges too far from `master`.

My concern with using `Any` is that if we did start to use the value then there would be no impetus to change the signature and we could easily end up with `Any`s persisting through the code. Using `object` forces it to be changed before it can be used, which is a good thing IMO.
If there are technical reasons not to use `object` in this way I'd really like to hear them because (as noted) I've started using this pattern elsewhere (for this reason) so would like to know what I've missed that makes this undesirable.

Regarding just removing this function -- I'd be happy with doing that if we think that that's a better solution here. Perhaps it is?

RealOrangeOne: I'm happy with `object` or removing the method, with a preference for the latter.

PeterJCLaw: 6a180a18a57558219b6c6a8dc97dc0d159e34596 removes this.

PeterJCLaw: (I'd still like to understand the original concern here though!)

RealOrangeOne (approved):

PR #209, opened 2020-11-05
Reduced timestep frequency

eljueves:


PeterJCLaw:
Thanks for putting this together.

Have you been able to determine that this does indeed result in the improvement we were looking for -- that it reduces the frequency, impact or both of the physics issues that we had last year? If so, please could you detail the testing you've done in the PR description; both so others can easily reproduce this on other machines and for future reference.

Separately, it might be worth rebasing this branch to remove the redundant commits? (I'd recommend doing that sort of thing prior to putting up a PR in future).

eljueves:
@PeterJCLaw I am not exactly sure what kind of physics problems that were occurring last year. I ran a couple simulations with 16ms time step and it was hard to tell the difference on my machine. do you have any scenarios that I can run?

eljueves:
I will rebase all PRs going forward. thanks for letting know

RealOrangeOne:
@eljueves the fact it's not obviously worse is definitely a good thing!

The main issues we saw were around robots crashing into eachother, especially with grabbers. When they do, it would often result in one of the robots being thrown violently from the arena.

eljueves:
@RealOrangeOne I will try that scenario

eljueves:
I'm going to close this pr and make a new one later

PR #212, opened 2020-11-08
Implement use_brake with torque/force control

OliverWales:
This MR implements the SR command `use_brake` to specify whether a motor should stop hard or coast when velocity is set to 0. It is implemented by modifying the available torque/force for rotary and linear motors respectively. The available force when coasting is set to 0.1 to provide some friction to have a nice rolling stop, but this value can be tuned in the future. It is a trade off between coasting nicely but often spinning vs staying under control but not coasting very far.

### DIFF
--- a/modules/sr/robot/motor.py
+++ b/modules/sr/robot/motor.py
@@ -82,4 +82,4 @@ class MotorChannel:
     @use_brake.setter
     def use_brake(self, value: bool):
         """Set whether to brake when motor speed is 0"""
-        self.sr_motor.set_brake(value)
+        self.sr_motor._set_brake(value)
--- a/modules/sr/robot/motor.py
+++ b/modules/sr/robot/motor.py
@@ -75,11 +75,11 @@ class MotorChannel:
         self.sr_motor.set_speed(translate(value, self.sr_motor))
 
     @property
-    def use_brake(self):
+    def use_brake(self) -> bool:
         """Whether to brake when motor speed is 0"""
         return self.sr_motor._use_brake
 
     @use_brake.setter
-    def use_brake(self, value):
+    def use_brake(self, value: bool):
         """Set whether to brake when motor speed is 0"""
         self.sr_motor.set_brake(value)
--- a/modules/sr/robot/motor.py
+++ b/modules/sr/robot/motor.py
@@ -80,6 +80,6 @@ class MotorChannel:
         return self.sr_motor._use_brake
 
     @use_brake.setter
-    def use_brake(self, value: bool):
+    def use_brake(self, value: bool) -> None:
         """Set whether to brake when motor speed is 0"""
         self.sr_motor._set_brake(value)
--- a/modules/sr/robot/motor.py
+++ b/modules/sr/robot/motor.py
@@ -50,8 +50,6 @@ class MotorChannel:
 
     def __init__(self, channel: int, sr_motor: Union[Gripper, Wheel, LinearMotor]) -> None:
         self.channel = channel
-        # Private shadow of use_brake
-        # self._use_brake = True # TODO create new thread for non-braking slowdown
 
         # There is currently no method for reading the power from a motor board
         self._power = 0
@@ -76,15 +74,12 @@ class MotorChannel:
 
         self.sr_motor.set_speed(translate(value, self.sr_motor))
 
-    ''''@property
-    def use_brake(self) -> bool:
-        "Whether to use the brake when at 0 speed"
-        return self._use_brake
+    @property
+    def use_brake(self):
+        """Whether to brake when motor speed is 0"""
+        return self.sr_motor._use_brake
 
     @use_brake.setter
-    def use_brake(self, value: bool) -> None:
-        self._use_brake = value
-
-        if self.power == 0:
-            "Implement the new braking setting"
-            self.power = 0'''
+    def use_brake(self, value):
+        """Set whether to brake when motor speed is 0"""
+        self.sr_motor.set_brake(value)
--- a/modules/sr/robot/motor_devices.py
+++ b/modules/sr/robot/motor_devices.py
@@ -13,6 +13,12 @@ class MotorBase:
         self.webot_motor = webot.getMotor(motor_name)
         self.max_speed = self.webot_motor.getMaxVelocity()
 
+        # Apply the brake when this motor speed = 0 by default
+        self._use_brake = True
+
+    def set_brake(self, value: bool) -> None:
+        self._use_brake = value
+
 
 class Wheel(MotorBase):
     """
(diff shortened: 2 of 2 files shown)


modules/sr/robot/motor.py
```diff
@@ -76,15 +74,12 @@ def power(self, value: int) -> None:
...
+        return self.sr_motor._use_brake
 
     @use_brake.setter
-    def use_brake(self, value: bool) -> None:
-        self._use_brake = value
-
-        if self.power == 0:
-            "Implement the new braking setting"
-            self.power = 0'''
+    def use_brake(self, value):
```
trickeydan, line 83:
```suggestion
    def _use_brake(self, value):
```

As we are not encouraging students to use this function directly, it's best that we make it a 'private' method.

modules/sr/robot/motor_devices.py
```diff
@@ -13,6 +13,12 @@ def __init__(self, webot: Robot, motor_name: str) -> None:
         self.webot_motor = webot.getMotor(motor_name)
         self.max_speed = self.webot_motor.getMaxVelocity()
 
+        # Apply the brake when this motor speed = 0 by default
+        self._use_brake = True
+
+    def set_brake(self, value: bool) -> None:
```
trickeydan, line 19:
```suggestion
    def _set_brake(self, value: bool) -> None:
```
As we are not encouraging students to call this function directly, we should make it 'private' by prefixing the function name with an underscore.

modules/sr/robot/motor_devices.py
```diff
@@ -23,10 +29,18 @@ def __init__(self, webot: Robot, motor_name: str) -> None:
...
         self.webot_motor.setPosition(float('inf'))
         self.webot_motor.setVelocity(0)
+        self.max_torque = self.webot_motor.getMaxTorque()
 
     def set_speed(self, speed: float) -> None:
         self.webot_motor.setVelocity(speed)
 
+        # If not using braking, set torque to 0.1 when motor speed is 0
+        # Otherwise set max torque
+        if(speed == 0 and not self._use_brake):
```
trickeydan, line 39:
```suggestion
        if speed == 0 and not self._use_brake:
```

Brackets not needed here :)

modules/sr/robot/motor_devices.py
```diff
@@ -46,6 +61,13 @@ def set_speed(self, speed: float) -> None:
             motor.setPosition(motor.getMaxPosition())
         motor.setVelocity(abs(speed))
 
+        # If not using braking, set force to 0.1 when motor speed is 0
+        # Otherwise set max force
+        if(speed == 0 and not self._use_brake):
```
trickeydan, line 66:
```suggestion
        if speed == 0 and not self._use_brake:
```
Same again with brackets here.

trickeydan (commented):
Looks good.

Might be a nice idea to reduce code duplication of the torque application into a `_apply_torque` function on the `MotorBase` class. This could then be called from `Wheel.set_speed` and `LinearMotor.set_speed` without affecting `Gripper`.

modules/sr/robot/motor_devices.py
PeterJCLaw: I disagree -- this class is already not part of the public API anyway, so I think it's more useful that we consider public/private at the level of "should this be called from outside the current class". In that context I think the current spelling is preferable.

If anything, I would encourage that `self._use_brake` be renamed to `self.use_brake` and just assigned to/read from directly (since we're already reading it directly).

modules/sr/robot/motor.py
```diff
@@ -76,15 +74,12 @@ def power(self, value: int) -> None:
 
         self.sr_motor.set_speed(translate(value, self.sr_motor))
 
-    ''''@property
-    def use_brake(self) -> bool:
-        "Whether to use the brake when at 0 speed"
-        return self._use_brake
+    @property
+    def use_brake(self):
```
PeterJCLaw, line 78: Not sure if your branch contains the latest from master -- we require type annotations everywhere, so we'll need one here. I think it should just be `bool`, but do check the docs.
(The setter below may also need one too).

modules/sr/robot/motor_devices.py
```diff
@@ -23,10 +29,18 @@ def __init__(self, webot: Robot, motor_name: str) -> None:
...
 
     def set_speed(self, speed: float) -> None:
         self.webot_motor.setVelocity(speed)
 
+        # If not using braking, set torque to 0.1 when motor speed is 0
+        # Otherwise set max torque
+        if(speed == 0 and not self._use_brake):
+            self.webot_motor.setAvailableTorque(0.1)
+        else:
+            self.webot_motor.setAvailableTorque(self.max_torque)
```
PeterJCLaw, line 42: As this is a method that we've not used before (also `setAvailableForce`), you may need to add it to the stubs in `stubs/controller.py`.

OliverWales: I think they are already in there - do I need to change them at all?

modules/sr/robot/motor_devices.py
OliverWales: Habits from other languages, will fix 👍

PeterJCLaw:
> Might be a nice idea to reduce code duplication of the torque application into a `_apply_torque` function on the `MotorBase` class. This could then be called from `Wheel.set_speed` and `LinearMotor.set_speed` without affecting `Gripper`.

Since the changes here involve different methods in each of `Wheel.set_speed` (which uses `setAvailableTorque`) and `LinearMotor.set_speed` (which uses `setAvailableForce`), I'm not sure there is duplication here?

Separately, why do you believe that this should not be applicable to `Gripper`?

modules/sr/robot/motor_devices.py
PeterJCLaw: Oh, if they're already there then likely they don't need changing. (Feel free to double check them though)
I was assuming that this might be the cause of the type checking errors you're seeing on CI, but I hadn't clicked through to actually look at the errors.

PeterJCLaw: Looks like I was wrong. I should have checked the errors first, sorry about that!

modules/sr/robot/motor.py
```diff
@@ -76,15 +74,12 @@ def power(self, value: int) -> None:
...
+        return self.sr_motor._use_brake
 
     @use_brake.setter
-    def use_brake(self, value: bool) -> None:
-        self._use_brake = value
-
-        if self.power == 0:
-            "Implement the new braking setting"
-            self.power = 0'''
+    def use_brake(self, value: bool):
```
PeterJCLaw, line 83: I think this needs a return type annotation:
```suggestion
    def use_brake(self, value: bool) -> None:
```

Issue #192, opened 2020-07-19
Validate and decrease timestep to reduce frequency of physics bugs


PeterJCLaw:
After some testing, we're fairly sure that we can reduce the rate of physics bugs by reducing the timestep from 32ms (the default) to 16ms.

While we don't want to do this for SR2020 (as it would likely affect the behaviour of the current robots), we should do this at some point.

When doing this we should validate that it does have the desired effect. See also #206 for other mitigations.

JennyLynnFletcher:
I would expect this change to significantly reduce the number of people able to run at 1.0x speed, so would be worth making ManualTimestepRobot as the default before reducing the basic timestep

eljueves:
Just made a pull request for this

PeterJCLaw:
We think https://github.com/srobo/competition-simulator/pull/230 sorted this another way.

PR #213, opened 2020-11-09
milestone #192: Validate and decrease timestep to reduce frequency of physics bugs

eljueves:
Reduced basicTimeStep of the world form 32 to 16. 

@RealOrangeOne I test the scenario that you told me about. I was not able to reproduce the issue at any time step 16 or 32. I did receive the following message in the webots console while testing both time stesps 

```
WARNING: The current physics step could not be computed correctly. Your world may be too complex. If this problem persists, try simplifying your bounding object(s), reducing the number of joints, or reducing WorldInfo.basicTimeStep.

```

I don't think it would be wise to go below 16ms


RealOrangeOne:
As a diff this obviously looks perfect, but it's more the testing.

The fact you didn't see it at all is concerning, but it required some pretty specific circumstances to happen. It's possible a mixture of updates and other changes have helped mitigate it.

Did you notice any changes in how demanding the simulations were to run? Was it still able to run at close to 1:1 time without pinning your computer?

eljueves:
@RealOrangeOne I did notice the robots handled a bit differently. Turning a robot left or right, for example, using the same power settings for the motors 32ms vs 16ms would lead the robot at slightly different angles

RealOrangeOne:
That can happen unfortunately. Was it by wildly different amounts or just a bit? So long as it's reliably the same between runs, and it's moving at a reasonable speed, I think that's ok.

eljueves:
@RealOrangeOne They were only off by a tiny amount. The speed of the simulation was not noticeably affected

RealOrangeOne:
Oh, then that's probably fine.

Why'd you close the PR?

PR #240, opened 2020-11-21
Some small tidyups in the territory controller

PeterJCLaw:
See commits for details.


WillB97 (approved):
The setup function was originally kept separate because of a concern that attempting to access the robot nodes during initialisation might cause issues. This appears to have been unfounded and this is a cleaner solution.

PR #242, opened 2020-11-26
Move territory radios to be level with robot radios

WillB97:
This makes the calculated distance to towers more accurate,
particularly at short ranges, since the api ignores the Y component.


prophile (approved):

RealOrangeOne (approved):
Nice find!

Issue #210, opened 2020-11-07
Automate running of matches


PeterJCLaw:
This aims to be a discussion post, as well as (eventually) a milestone target for some work to do in this area.

It would be great to automate the running of matches. This will reduce the amount of volunteer effort is needed to run a match session, ideally to the point where running all the matches for a given (league) session is as close to a one-click solution as we can reasonably get it. I'm deliberately ignoring knockouts here as that would introduce a dependency on Automated Scoring (see also #18), which I'd like to leave out of scope for now.

I'm therefore aiming to capture the things we want from competition matches so that we know what to automate. I hope that we can then also have a discussion about how best to achieve those outcomes, potentially in different ways than we have in the past.

## Requirements

1. Teams code needs to be put somewhere so that the right robot controller can run it
2. The simulated world should be fresh at the start of the match
3. The simulated match should be recorded (animation and video)
  a. ideally this would be tagged with the match id, for our convenience
4. Errors during the match should be reported clearly to the user
5. Teams log output needs to end up somewhere so that we can share it with them afterwards
  a. it should be clear to the team which log came from which match, and
  b. ideally this would group by team for our convenience

Note: I place to update the above list as we think of things -- please do add to it if there are useful things to add.

PeterJCLaw:
## Existing solutions
(as of 2020-11-07)

1. Currently we have a `prepare-comp-match` script which handles putting the team code in the right places.
2. We assume that the simulation is started afresh each time from a clean instance of the repo
3. When run in `comp` mode we automatically capture video and animation
  a. no solution for match ids yet :(
4. We rely on the user spotting errors in Webot's console :(
5. Robot logs end up next to the respective code
  a. no solution for match ids yet :(
  b. being next to the code sort-of allows this, but it's not the easiest to map back to TLAs :(

#185 aims to fix 5. b., however ends up quite coupled to SRComp internals and doesn't solve 5. a.

Otherwise we currently handle things manually.

PeterJCLaw:
## Suggestion for future direction

I think we would benefit from moving towards explicit configuration of the simulator in `comp` mode. At the moment we have a number of aspects which work separately to achieve parts of this, but which I think we should centralise.
I would still like it to be possible to control each of the individually (certainly I _don't_ want `comp` mode to become a single switch which is the only way to control lots of separate behaviours), even if `comp` mode does change a lot of defaults.

[...]

PR #243, opened 2020-11-26
Extract controller utils

PeterJCLaw:
This is better than cross-importing between controllers and provides a single place for additional utils to be placed.

This is partly in preparation for other changes towards #210.


OliverWales (approved):
Looks good to me

Issue #189, opened 2020-07-19
Automate collation of logs by team


PeterJCLaw:
We have scripts to organise the start of a match based on the teams on the match and the match number and we want to be able to share the logs per match with teams after the broadcast.

It would be great if we had a way to collate the logs per team as we go, in order to simplify the job of doing the recordings.

PeterJCLaw:
Likely we want to update the match preparation script to record which teams it's set up for the match and then use that data to drive where the logs go.

PeterJCLaw:
A solution I though about for this would be to have the supervisor controller doing this after the match is paused.

PeterJCLaw:
#256 does this in the runner script.

Issue #18, opened 2020-05-06
Live scoring


antoinepetty:
Implement a way to score each round automatically (and display this on screen?)

This will involve at least:
- checking that we can detect cases where an object is "inside" or "touching" another object (at least in the 2D arena sense, possibly as well as implementing the three-corner rule or similar)
- recording changes to that state, either as a log of change events or as snapshots of the new state on change, or similar so that we can replay the scoring later; likely we'd want match-timestamps here to help line up to the animation

I suggest that we probably want a separate Supervisor to handle the scoring, if Webots supports having more than one.
Note that this scoring Supervisor may need to communicate with the arena Supervisor (see #216), so it may make sense for them to be the same process.

PeterJCLaw:
Tying the match running to SRComp's HTTP API would likely be somewhat complicated on a couple of fronts:
- SRComp is designed around consuming fixed data from a git repo, so writing data to it via the API would be quite a big change. Not sure if that's what you had in mind though
- for SR2020 we ended up actually running the simulations of the matches slightly ahead of the broadcast in order to ensure that they worked as expected and for our own organisational simplicity; we're unlikely to change this, meaning that the API (at least the public facing one) wouldn't necessarily actually be in sync with the matches being run

Instead I think we're more likely to head down the route of having the supervisor "robot" doing a bunch of the match coordination side of things and then outputting a structured log (with timestamps) of what happened so we can feed that through the system.

There is of course plenty of room to feed those output logs into SRComp to expose the information to API consumers (would need work in SRComp, however as an extension to the existing match score files I can see this fitting reasonably well).

PeterJCLaw:
Note for picking this up -- I suspect that this might be easier after we've fixed #83 since that also involves interpreting the positions of items in the world.

WillB97:
The sr2021 game has a secondary supervisor running the territory_controller. Currently each time a zone is claimed a line is printed to stdout with the station code, claimant and simulation time. It just needs to be converted into a form that srcomp can accept.

PR #244, opened 2020-11-28
Record match & scoring data

PeterJCLaw:
This has the competition mode runs of the simulator output a JSON file which is compatible with the [Proton](https://github.com/PeterJCLaw/proton) specification used by SRComp for scoring.

This will also allow us to more easily move the robot logs to suitable directories for storage, though that is not addressed here.

This contributes towards #18, #210 and #189.

Suggest review by commit.

Note: this may end up conflicting with #240, though hopefully not too badly.

See https://github.com/srobo/sr2021-comp/pull/1 for the scorer implementation.

This is part 1 of 3 towards match automation:
- Record match & scoring data (this PR #244)
- Add recording controls (#254)
- Automate running of individual matches (#256)


script/prepare-comp-match
```diff
@@ -14,8 +14,7 @@ REPO_ROOT = Path(__file__).resolve().parent.parent
 sys.path.insert(1, str(REPO_ROOT / 'modules'))
 
 import controller_utils  # isort:skip
-
-NUM_ZONES = 2
+from controller_utils import NUM_ZONES  # isort:skip
```
RealOrangeOne, line 17: This seems unneessary, can you not just replace the uses of `NUM_ZONES` with `controller_utils.NUM_ZONES`?

PeterJCLaw: For sure we could, this felt like it left that code slightly clearer though (and avoided some slightly odd wrapping).

PeterJCLaw: 0dfc03781cc75adb159aab5bceb513c618ce86ac

WillB97 (approved):
Once #256 and #254 have been merged.

PeterJCLaw:
> Once #256 and #254 have been merged.

Why couple those changes to this? This PR doesn't depend on those (somewhat deliberately so); rather they depend upon this PR. While these changes were designed with those subsequent changes in mind, they were also designed to be useful and sufficient on their own.

WillB97:
I was incorrect, but #256 is needed since the excess logging I/O which was fixed there appears to effect this PR as well.

PeterJCLaw:
> I was incorrect, but #256 is needed since the excess logging I/O which was fixed there appears to effect this PR as well.

Ah, I see. I'd rather address that by pulling the I/O improvements here and then merging. Coupling this PR to those just for that fix doesn't seem like a great solution to me. I'll do some branch juggling shortly to arrange that.

PR #246, opened 2020-12-03
Add a few scientific libraries

sedders123:
As discussed in the latest competition meeting we want to add these libraries for teams to use.

I had to change our requirement checking logic to use regex to identify packages instead of partitioning strings due to the extra pip functionality required by pytorch in `libraries.txt`. I've attempted to make [the regex](regexr.com/5hk94) suitably generic so that the `get_distribution` call will work across operating systems (specifically `+cpu` wont be there on no win32 platforms). Its almost certainly not perfect but should work for the packages we currently depend on.

### DIFF
--- a/controllers/competition_supervisor/competition_supervisor.py
+++ b/controllers/competition_supervisor/competition_supervisor.py
@@ -1,4 +1,3 @@
-import re
 import sys
 import time
 import datetime
@@ -72,10 +71,9 @@ def check_required_libraries(path: Path) -> None:
     missing, incorrect = [], []
 
     for package in path.read_text().splitlines():
-        package_match = re.match(r"[\w|-]+==[\d|.]+", package)
-        if not package_match:
+        package = package.partition('#')[0].strip()
+        if not package:
             continue
-        package = package_match.group(0)
 
         try:
             pkg_resources.get_distribution(package)
--- a/controllers/competition_supervisor/competition_supervisor.py
+++ b/controllers/competition_supervisor/competition_supervisor.py
@@ -1,3 +1,4 @@
+import re
 import sys
 import time
 import datetime
@@ -71,9 +72,10 @@ def check_required_libraries(path: Path) -> None:
     missing, incorrect = [], []
 
     for package in path.read_text().splitlines():
-        package = package.partition('#')[0].strip()
-        if not package:
+        package_match = re.match(r"[\w|-]+==[\d|.]+", package)
+        if not package_match:
             continue
+        package = package_match.group(0)
 
         try:
             pkg_resources.get_distribution(package)
--- a/libraries.txt
+++ b/libraries.txt
@@ -11,10 +11,5 @@
 matplotlib==3.3.3
 numpy==1.19.3
 pandas==1.1.4
-torch==1.7.0 ; sys_platform != "win32"
 scikit-learn==0.23.2
 scipy==1.5.4
-
-
--f https://download.pytorch.org/whl/torch_stable.html
-torch==1.7.0+cpu ; sys_platform == "win32"
--- a/libraries.txt
+++ b/libraries.txt
@@ -8,4 +8,12 @@
 # matches, so we *don't* have the full range of requirements.txt supported
 # syntax available.
 
-matplotlib==3.2.2
+matplotlib==3.3.3
+numpy==1.19.3
+pandas==1.1.4
+torch==1.7.0 ; sys_platform != "win32"
+scipy==1.5.4
+
+
+-f https://download.pytorch.org/whl/torch_stable.html
+torch==1.7.0+cpu ; sys_platform == "win32"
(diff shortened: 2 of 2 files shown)


.vscode/settings.json
```diff
@@ -2,5 +2,5 @@
     "python.autoComplete.extraPaths": ["modules", "stubs"],
     "python.linting.flake8Enabled": true,
     "python.linting.mypyEnabled": true,
-    "python.pythonPath": "${env:VIRTUAL_ENV}/bin/python",
+    "python.pythonPath": "C:\\Python38\\python.exe",
```
RealOrangeOne, line 5: This should probably be reverted

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -71,9 +72,10 @@ def check_required_libraries(path: Path) -> None:
     missing, incorrect = [], []
 
     for package in path.read_text().splitlines():
-        package = package.partition('#')[0].strip()
-        if not package:
+        package_match = re.match(r"[\w|-]+==[\d|.]+", package)
```
RealOrangeOne, line 75: I wonder if we'd be better off using the actual parser for this?

https://stackoverflow.com/questions/49689880/proper-way-to-parse-requirements-file-after-pip-upgrade-to-pip-10-x-x mentions that `pip._internal.parse_requirements` does the job, and from some local testing with our file it does (`pkg_resources.parse_requirements` doesn't like the `-f` option).

RealOrangeOne:
I've also noticed we don't attempt to install the `libraries.txt` file on CI. It might be worth doing that, mostly to check it parses correctly.

libraries.txt
```diff
@@ -8,4 +8,13 @@
 # matches, so we *don't* have the full range of requirements.txt supported
 # syntax available.
 
-matplotlib==3.2.2
+matplotlib==3.3.3
+numpy==1.19.3
+pandas==1.1.4
+torch==1.7.0 ; sys_platform != "win32"
```
PeterJCLaw, line 14: PyTorch is _enormous_ (~1.5G if memory serves). Do we really want to force everyone who develops the simulator to need to download a huge binary, especially as the majority of users aren't going to actually be using it?

If the objection to TensorFlow was size and portability issues, I'm not sure how PyTorch solves that? (It's bigger and similarly has native features which may or not behave the same across platforms)

Is there an alternative mechanism which competitors who want to do machine learning could use, but which doesn't require imposing such limitations on others?

PeterJCLaw:
Should we consider pinning dependencies all the way down? Now that we've got lots of things it might help clarify what that actually gets you. [`pip-tools`](https://github.com/jazzband/pip-tools) `pip-compile` command is very useful for that -- we can specify a `libraries.in` file (which would essentially be what we have as `libraries.txt` at the moment) and it can generate a `libraries.txt`. It does have the disadvantage that the `libraries.txt` it generates is platform specific though.

Some notable libraries which are actually present but not listed:
- `pillow`
- `python-dateutil`
- `typing-extensions`
- `kiwisolver`

libraries.txt
sedders123: @prophile is there a reason we thought PyTorch was better than TensorFlow, I can't remember what we said in the meeting

RealOrangeOne: It was assumed pytorch was smaller than tensorflow, and much easier to install.

PeterJCLaw: I'm pretty sure that they both ship binary wheels (at least on linux), so it should just be `pip install ...`. PyTorch looks like it's variable in size across platforms, which _may_ suggest that there are some other things needed on other platforms?

controllers/competition_supervisor/competition_supervisor.py
sersorrel: I would strongly recommend against using pip's internals to do this: https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program (apart from anything else, if competitors are going to be using this, it's likely to run against many different versions of pip, and so more likely that someone will run into a pip API change that breaks it)

Could you do something like remove lines that look like arguments to pip (e.g. lines that start with `-`) before passing them to `pkg_resources`?

Regardless: in character classes (`[...]`), pipes are interpreted literally, so currently this would match a line like `foo|bar==12|3`.

libraries.txt
RealOrangeOne: In meeting last night, we concluded that chances are either PyTorch or Tensorflow are probably massively overkill for this, and something like scikit-learn would likely cover everything needed, and be much more useful. Not to mention the install is significantly smaller, and thus makes development much easier.

So let's remove PyTorch.

PeterJCLaw: https://github.com/srobo/competition-simulator/pull/246/commits/c99e323c274c9e055be6374204a581d390cb039c

PeterJCLaw (approved):

PR #247, opened 2020-12-05
Simulator tweaks - Video improvements

WillB97:
Reduces the timestep to 8 since this appears to have no detriment to performance and changes the movie frame rate to 25 from 31.25.
Also sets comp mode simulation mode to run since we no longer need to run at real-time speed so we can save some time running matches.

### DIFF
--- a/controllers/competition_supervisor/competition_supervisor.py
+++ b/controllers/competition_supervisor/competition_supervisor.py
@@ -145,7 +145,7 @@ def run_match(supervisor: Supervisor) -> None:
         robot.getField('customData').setSFString('start')
 
     # ... then un-pause the simulation, so they all start together
-    supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_REAL_TIME)
+    supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_RUN)
 
     time_step = int(supervisor.getBasicTimeStep())
     duration_ms = time_step * int(1000 * GAME_DURATION_SECONDS // time_step)
--- a/worlds/Arena.wbt
+++ b/worlds/Arena.wbt
@@ -1,7 +1,7 @@
 #VRML_SIM R2020b utf8
 WorldInfo {
   coordinateSystem "NUE"
-  basicTimeStep 16
+  basicTimeStep 8
 }
 Viewpoint {
   orientation -0.9999095454609961 -0.012169594624247947 0.0057272910407683165 0.8997040656177113


controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -145,7 +145,7 @@ def run_match(supervisor: Supervisor) -> None:
         robot.getField('customData').setSFString('start')
 
     # ... then un-pause the simulation, so they all start together
-    supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_REAL_TIME)
+    supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_RUN)
```
Adimote, line 148: possibly a controversial change? I'm worried teams might use normal `time` instead of `R.time` to measure timings, and thus if we run it at faster-than-realtime it might cause issues. We'd need to inform teams more clearly about the behavior of the simulation if we ship this.

worlds/Arena.wbt
```diff
@@ -1,7 +1,7 @@
 #VRML_SIM R2020b utf8
 WorldInfo {
   coordinateSystem "NUE"
-  basicTimeStep 16
+  basicTimeStep 8
```
Adimote, line 4: 👍 from my experiments earlier this didn't seem to change robot behavior visibly, may need to warn teams that it's happening though

controllers/competition_supervisor/competition_supervisor.py
Adimote: Actually, I take it back. It seems 0 teams used `time.sleep()` in the finals last year

Adimote (approved):
Approved, but we should probably warn teams the timestep is changing

controllers/competition_supervisor/competition_supervisor.py
WillB97: Also I've noticed that real-time is an approximation that swings between 0.9x and 1.1x so teams using time.sleep are going to be having issues anyway

worlds/Arena.wbt
RealOrangeOne: When we publish updates we should also publish a changelog anyway in GitHub releases, so that covers that I think.

RealOrangeOne (approved):
I think this looks like it should be fine. Webots do recommend the timestep being as low as possible.

My laptop seems to run the simulation about 4.5x, so I think we're find for performance, which was my main concern with the changes.

We should definitely mention both the timestep change and that the comp simulation will be run as fast as possible. There might be some documentation changes around how to run a simulation, but we can address that when releasing.

PR #249, opened 2020-12-16
Fix depreciation warnings in webots 2021a

WillB97:
This pull request currently only replaces the depreciated `getX` methods with the `getDevice` method.
While this works in webots 2021a the `getDevice` method is [broken in 2020b](https://github.com/cyberbotics/webots/issues/2338) so fallbacks will be needed.
Additionally the stubs and types need to be added.

TLDR before merging:
- [ ] Add method stub for `Robot.getDevice`
- [ ] Add fallbacks to use previous methods for webots-202b compatibility
- [ ] Satisfy the typechecking for `getDevice` in all it's uses

### DIFF
--- a/controllers/territory_controller/territory_controller.py
+++ b/controllers/territory_controller/territory_controller.py
@@ -57,10 +57,10 @@ class TerritoryController:
         return self._station_statuses[station_code]
 
     def setup(self) -> None:
-        self._emitters = {station_code: self._robot.getEmitter(station_code + "Emitter")
+        self._emitters = {station_code: self._robot.getDevice(station_code + "Emitter")
                           for station_code in StationCode}
 
-        self._receivers = {station_code: self._robot.getReceiver(station_code + "Receiver")
+        self._receivers = {station_code: self._robot.getDevice(station_code + "Receiver")
                            for station_code in StationCode}
         territory_controller.enable_receivers()
 
--- a/modules/sr/robot/ruggeduino_devices.py
+++ b/modules/sr/robot/ruggeduino_devices.py
@@ -17,7 +17,7 @@ class DistanceSensor:
     UPPER_BOUND = 2
 
     def __init__(self, webot: Robot, sensor_name: str) -> None:
-        self.webot_sensor = webot.getDistanceSensor(sensor_name)
+        self.webot_sensor = webot.getDevice(sensor_name)
         self.webot_sensor.enable(int(webot.getBasicTimeStep()))
 
     def __get_scaled_distance(self) -> float:
@@ -46,7 +46,7 @@ class Microswitch:
     """
 
     def __init__(self, webot: Robot, sensor_name: str) -> None:
-        self.webot_sensor = webot.getTouchSensor(sensor_name)
+        self.webot_sensor = webot.getDevice(sensor_name)
         self.webot_sensor.enable(int(webot.getBasicTimeStep()))
 
     def read_value(self) -> bool:
@@ -69,7 +69,7 @@ class Led:
         limiter: OutputFrequencyLimiter,
     ) -> None:
         self._name = device_name
-        self.webot_sensor = webot.getLED(device_name)
+        self.webot_sensor = webot.getDevice(device_name)
         self._limiter = limiter
 
     def write_value(self, value: bool) -> None:
--- a/modules/sr/robot/motor_devices.py
+++ b/modules/sr/robot/motor_devices.py
@@ -10,7 +10,7 @@ class MotorBase:
 
     def __init__(self, webot: Robot, motor_name: str) -> None:
         self.motor_name = motor_name
-        self.webot_motor = webot.getMotor(motor_name)
+        self.webot_motor = webot.getDevice(motor_name)
         self.max_speed = self.webot_motor.getMaxVelocity()
 
 
--- a/modules/sr/robot/radio.py
+++ b/modules/sr/robot/radio.py
@@ -92,9 +92,9 @@ class Radio:
 
     def __init__(self, webot: Robot, zone: int, step_lock: Lock) -> None:
         self._webot = webot
-        self._receiver = webot.getReceiver("robot receiver")
+        self._receiver = webot.getDevice("robot receiver")
         self._receiver.enable(1)
-        self._emitter = webot.getEmitter("robot emitter")
+        self._emitter = webot.getDevice("robot emitter")
         self._zone = zone
         self._step_lock = step_lock
 


WillB97:
I'm not sure how to handle getDevice() being assignable to Optional[Emitter], Optional[Receiver], Optional[Motor], Optional[LED], Optional[DistanceSensor] or Optional[TouchSensor] since setting the return type to a union of all of them still appears to fail typechecking.

modules/sr/robot/utils.py
```diff
@@ -7,3 +20,75 @@ def map_to_range(
...
+
+def get_robot_emitter(robot: Robot, device: str) -> Emitter:
+    try:
+        # webots 2020b fails to assign the output to the expected type
+        emitter: Optional[Emitter] = robot.getDevice(device)
+        # webots 2020b always returns None when not raising TypeError
+        if emitter is None:
+            raise TypeError
+    except TypeError:
+        emitter = robot.getEmitter(device)
```
PeterJCLaw, line 33: This pattern looks very odd -- probably worth a comment about why we're calling `getEmitter` on `TypeError`.

Are there other sources of `TypeError` which this is catching, or is this just for our own one? If the latter, then I suggest simple branching might be neater. Otherwise a comment about the other source might be useful.

modules/sr/robot/utils.py
```diff
@@ -7,3 +20,75 @@ def map_to_range(
 ) -> float:
     """Maps a value from within one range of inputs to within a range of outputs."""
     return ((value - old_min) / (old_max - old_min)) * (new_max - new_min) + new_min
+
+
+def get_robot_emitter(robot: Robot, device: str) -> Emitter:
+    try:
+        # webots 2020b fails to assign the output to the expected type
+        emitter: Optional[Emitter] = robot.getDevice(device)
```
PeterJCLaw, line 28: Regarding how to handle the type errors you'll be getting here -- we need to either just plain [`cast`](https://docs.python.org/3/library/typing.html#typing.cast) the object, or check it first and then assign it (maybe also with a cast).

I would suggest that we go with an explicit check, in part because we could then extract our own helper function which looks something like this:
``` python
form typing import Type, TypeVar
from controller import Device

TDevice = TypeVar('TDevice', bound=Device)

def get_robot_device(robot: Robot, device: str, kind: Type[TDevice]) -> TDevice:
    device: Optional[Device] = robot.getDevice(device)
    if not isinstance(device, kind):
        raise TypeError(...)
    return cast(TDevice, device)
```
The trick here is that we're using `Type[T] -> T`  to allow us to operate generically over all possible `Device` types while preserving the type information the caller has provided.
Usage looks like this:
``` python
emitter: Emitter = get_robot_device(robot, 'emitter', Emitter)
```
I've omitted the fallback interactions above as I'm not sure how they behave. Hopefully there's enough information about the types above to get you going though.

modules/sr/robot/utils.py
```diff
@@ -7,3 +20,75 @@ def map_to_range(
...
+            raise TypeError
+    except TypeError:
+        distance_sensor = robot.getDistanceSensor(device)
+    return distance_sensor
+
+
+def get_robot_touch_sensor(robot: Robot, device: str) -> TouchSensor:
+    try:
+        # webots 2020b fails to assign the output to the expected type
+        touch_Sensor: Optional[TouchSensor] = robot.getDevice(device)
```
PeterJCLaw, line 88: nit: `s/touch_Sensor/touch_sensor/`.

controllers/territory_controller/territory_controller.py
```diff
@@ -57,10 +66,12 @@ def get_claimant(
         return self._station_statuses[station_code]
 
     def setup(self) -> None:
-        self._emitters = {station_code: self._robot.getEmitter(station_code + "Emitter")
+        self._emitters = {station_code:
+                          get_robot_emitter(self._robot, station_code + "Emitter")
                           for station_code in StationCode}
```
PeterJCLaw, line 71: nit: I think I'd expect this to be wrapped like:
```suggestion
        self._emitters = {
            station_code: get_robot_emitter(self._robot, station_code + "Emitter")
            for station_code in StationCode
        }
```
I realise that's not quite like the other comprehensions in this file, though that's a result of this file being out of step with the rest of this repo. (Unfortunately we don't have a linter for this yet)

(also `_receivers` below)

PeterJCLaw (commented):
Various comments, hopefully there's enough there around the types to unblock you on that. Happy to talk through that if you get stuck though -- I know generics aren't the easiest thing to get your head around!

modules/sr/robot/utils.py
WillB97: getDevice seems to have rather odd behaviour in 2020b. On the one hand assigning the output to an untyped variable makes the variable Nonetype while assigning to type Optional[Emitter] throws an internal type error.
```
1|   File ".../competition-simulator/modules/sr/robot/motor_devices.py", line 14, in __init__
1|     self.webot_motor = get_robot_motor(webot, motor_name)
1|   File ".../competition-simulator/modules/sr/robot/utils.py", line 52, in get_robot_motor
1|     motor: Optional[Motor] = robot.getDevice(device)
1|   File "/usr/local/webots/lib/controller/python38/controller.py", line 2640, in getDevice
1|     if size == 0 or tag >= size:
1| TypeError: '>=' not supported between instances of 'str' and 'int'
```
Since I can't guarantee that the error will always occur I was just covvering my bases.
I'll add the type error information to the comment at the start of the try statement.

modules/sr/robot/utils.py
WillB97: Since each fallback function is specific to the type being returned they could be added with an if-else tree based on the kind argument.
This method seems like a better solution since all the current functions perform aproximately the same actions.

WillB97: I have a strange issue where on webots-2020b robot.getDevice throws a TypeError but if I wrap that line in a try/except and put the fallback function in the error handler the fallback function isn't executed and the instance test fails.

modules/sr/robot/utils.py
PeterJCLaw: I think it's worth thinking about the code flow as separate from the type annotations. In general type annotations don't have any effect at runtime, so we ought to see the same behaviour regardless of the annotations (if we're not then something very funky is going on).

It looks like in `2020b` the code in `getDevice` is expecting an `int`:
``` python
    def getDevice(tag):
      if tag == 0:
          return None
      size = len(Robot.__devices)
      if size == 0 or tag >= size:
          return None
      return Robot.__devices[tag]
```
What does the code look like in 2021a?

----
I think it might help clarity if we condense the `try...except TypeError` around just the part of the code which we know will raise `TypeError` and move our own `None` check outside that. If we're primarily catching `TypeError` to work around the version differences then it may make sense to have our own helper which does literally just the version differences handling (but none of the type annotation handling) which we can then wrap in handlers which do the type specific stuff.

modules/sr/robot/utils.py
WillB97: This works but removing the try except throws a TypeError and merging the except and device is None clauses causes that code to never be run.
```python
def get_robot_device(robot: Robot, name: str, kind: Type[TDevice]) -> TDevice:
    device: Optional[Device] = None
    try:
        device = robot.getDevice(name)
    except TypeError:
        pass
    if device is None:
        if kind is Motor:
            device = robot.getMotor(name)
    if not isinstance(device, kind):
        raise TypeError
    return cast(TDevice, device)
```

modules/sr/robot/ruggeduino_devices.py
```diff
@@ -1,7 +1,12 @@
 import logging
 
-from controller import Robot
-from sr.robot.utils import map_to_range
+from controller import (
+    LED,
+    Robot,
+    TouchSensor,
+    DistanceSensor as TDistanceSensor,
```
PeterJCLaw, line 7: I would encourage not using `T` as the prefix here -- that tends to imply that this is a generic type parameter, which I don't think that this is. Perhaps `WebotsDistanceSensor`?

WillB97: Yeah `WebotsDistanceSensor` is a better name hopefully this will be superseded by #250

modules/sr/robot/utils.py
```diff
@@ -7,3 +21,30 @@ def map_to_range(
 ) -> float:
     """Maps a value from within one range of inputs to within a range of outputs."""
     return ((value - old_min) / (old_max - old_min)) * (new_max - new_min) + new_min
+
+
+TDevice = TypeVar('TDevice', bound=Device)
```
PeterJCLaw, line 26: Small thing: I'd encourage that type definitions appear just below imports and above all function definitions and/or constants.

stubs/controller.py
```diff
@@ -222,6 +225,8 @@ def getMotor(self, name: str) -> Motor: ...
     def getReceiver(self, name: str) -> Receiver: ...
     def getTouchSensor(self, name: str) -> TouchSensor: ...
 
+    def getDevice(self, name: str) -> Optional[TDevice]: ...
```
PeterJCLaw, line 228: I'm surprised this works? I think this should just be `Device` rather than `TDevice`. For generic parameters in return types there generally needs to be another mention of the same type parameter within the types of the arguments -- so that you can know what the concrete type is when the function is called.

modules/sr/robot/utils.py
```diff
@@ -7,3 +21,30 @@ def map_to_range(
...
+
+
+def get_robot_device(robot: Robot, name: str, kind: Type[TDevice]) -> TDevice:
+    device: Optional[Device] = None
+    try:  # webots 2020b fails to assign the output to the expected type raising TypeError
```
PeterJCLaw, line 31: I'm not sure this comment is quite correct. Perhaps let's just describe this as:
```suggestion
    try:
        # webots 2020b is buggy and always TypeError, however we're aiming to
        # be forwards compatible with 2021a, so try this first.
```

modules/sr/robot/utils.py
WillB97: 2021a seems to run similar code
```python
%rename ("__internalGetDeviceTagFromName") webots::Robot::getDeviceTagFromName;
%extend webots::Robot {
  %pythoncode %{
def getDevice(self, name):
  tag = self.__internalGetDeviceTagFromName(name)
  return self.__getOrCreateDevice(tag)
@staticmethod
def internalGetDeviceFromTag(tag):
  if tag == 0:
      return None
  size = len(Robot.__devices)
  if size == 0 or tag >= size:
      return None
  return Robot.__devices[tag]
```
I've changed the error handling code to the recommendation in the comment below

PeterJCLaw: Note the call to `__internalGetDeviceTagFromName` to convert the incoming parameter to `getDevice` from a `name` into a `tag` though -- I bet that's the fix.

Your new code looks good though :+1:. (I'll have a better look this evening)

PeterJCLaw (commented):
Tested locally in `R2020b revision 1`, this works. I don't have 2021a to hand to test with..

PeterJCLaw (approved):

PR #251, opened 2020-12-18
Fixes comp mode crashing in webots-2021a

WillB97:
Version 2021a removed the RUN mode from the simulation mode enum which causes webots to crash if this value is used.
The replacement value FAST has different properties in version 2020b causing comp mode to fail to record, so a workaround is used to detect which version is running and use the appropriate value.

The version detection is a hack using the changed behaviour of getDevice between versions.


controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -145,7 +145,11 @@ def run_match(supervisor: Supervisor) -> None:
         robot.getField('customData').setSFString('start')
 
     # ... then un-pause the simulation, so they all start together
-    supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_RUN)
+    if supervisor.getDevice("2021a-compatibility") is None:
+        # webots-2021a removed the RUN mode and now uses FAST
```
RealOrangeOne, line 149: Given this comment, are these the right way around?

worlds/Arena.wbt
```diff
@@ -41,6 +41,13 @@ Robot {
   description "A supervisor 'robot' to help run the competition."
   controller "competition_supervisor"
   supervisor TRUE
+  children [
+    # this is a hack to detect if we are running on version 2020b
```
RealOrangeOne, line 45: So what's the idea here? Have a useless generic device which can be retrieved by `getDevice`, simply to check it returns null?

Could this name be added to the supervisor itself? Save adding an extra node?

controllers/competition_supervisor/competition_supervisor.py
WillB97: More likely the comment is unclear. If 2021a-compatibility is None, you are running 2020b

worlds/Arena.wbt
WillB97: I did try that but in order for `getDevice` to return something on 2021a you need a node that subclasses `Device` that is a child of the robot the controller is running on.

PeterJCLaw:
Meta: if 2021a has these differences, do we really want to support it in `comp` mode? Might it be better instead to have it fail (ideally gracefully) to tell people that their environment doesn't match what the competition will be run as?

WillB97:
Since the current plan is to run 2020b for league 1 and then 2021a after that, we could change this PR to just fail if it detects 2021a and then flip the logic after league 1 to pin it to 2021a+.
The extra node hack would still be needed to detect the version so we can fail gracefully.

PeterJCLaw:
> Since the current plan is to run 2020b for league 1 and then 2021a after that, we could change this PR to just fail if it detects 2021a and then flip the logic after league 1 to pin it to 2021a+.
> The extra node hack would still be needed to detect the version so we can fail gracefully.

This perhaps isn't the place to discuss this, but I don't think we should change the reference environment during the competition year.

PeterJCLaw:
ahgh, sorry, that's not the button I wanted.

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -145,7 +145,12 @@ def run_match(supervisor: Supervisor) -> None:
         robot.getField('customData').setSFString('start')
 
     # ... then un-pause the simulation, so they all start together
-    supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_RUN)
+    if supervisor.getDevice("2021a-compatibility") is None:
+        # we are running version 2020b so the old command is used
+        supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_RUN)
+    else:
+        # webots-2021a removed the RUN mode and now uses FAST
+        supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_FAST)
```
WillB97, line 153: Do we actually want to prevent curious teams from being able to run an actual match because they are using a newer version?
Could we just output a clear warning that their environment doesn't match the competition setup, such as:
```suggestion
        print("All offical competition matches will be run on version 2020b-rev1 unless otherwise stated", file=sys.stderr)
        print("Since this simulator is running a different version some behaviour may not match that of the offical competition matches", file=sys.stderr)
        supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_FAST)
```

sedders123: Minor tweaks and typo fixes
>All official competition matches will be run on Webots version 2020b-rev1 unless otherwise stated
>
>Since this simulator is running a different version it is possible that some behaviour may not match that of the official competition matches

WillB97: That does read better so I'm happy to use that

RealOrangeOne: Personally i'd rather we mention the competition version in the docs rather than the simulator logs. It makes things easier to change in future, and keeps the logs clean.

Also, are we sure the _competition_ will run 2020b-rev1? I thought we had decided the first league session would, and we'd keep an eye from there?

controllers/competition_supervisor/competition_supervisor.py
WillB97: The comments are changed in 8873fdc54fb38a5f3fb498eed09181bc475a72af

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -145,7 +145,23 @@ def run_match(supervisor: Supervisor) -> None:
...
+            "Webots version 2020b-rev1 unless otherwise stated",
+            file=sys.stderr,
+        )
+        print(
+            "Since this simulator is running a different version "
+            "it is possible that some behaviour may not match "
+            "that of the official competition matches",
+            file=sys.stderr,
+        )
+        supervisor.simulationSetMode(Supervisor.SIMULATION_MODE_FAST)
```
PeterJCLaw, line 164: Would be great to pull this out to it's own util. The rest of this function is fairly concentrated on just the things needed to actually run the match, yet this block has quite a lot of detail which isn't really relevant to that.

This block ending up something like this might be good?
```python
    supervisor.simulationSetMode(get_simulation_run_mode())
```
Alternatively just having this block within a function like this would be good:
```python
    set_simulation_run_mode(supervisor)
```

controllers/competition_supervisor/competition_supervisor.py
WillB97: > Personally i'd rather we mention the competition version in the docs rather than the simulator logs. It makes things easier to change in future, and keeps the logs clean.

We could reduce the warning to just:
```
This simulator is running a different version than is being used for the next official competition matches, please check the docs for which version will be used
As such it is possible that some behaviour may not match that of the official competition matches
``` 

> Also, are we sure the _competition_ will run 2020b-rev1? I thought we had decided the first league session would, and we'd keep an eye from there?

Agreed we could do with being more vague about this to give ourselves options

controllers/competition_supervisor/competition_supervisor.py
WillB97: > Would be great to pull this out to it's own util. The rest of this function is fairly concentrated on just the things needed to actually run the match, yet this block has quite a lot of detail which isn't really relevant to that.

We could pull this out into a function but it should remain within the competition_supervisor since it operates directly upon the supervisor's node.

> This block ending up something like this might be good?
> 
> ```python
>     supervisor.simulationSetMode(get_simulation_run_mode())
> ```

I prefer this format since the overall parameter can be referred to as the run mode so a setter is quite ambiguous.

controllers/competition_supervisor/competition_supervisor.py
sedders123: I like the reduced warning. Below is how I would personally tweak it but feel free to take or leave any changes as this is getting quite nit picky.
```
This simulator is running a different version of Webots to the one that will be used for the next official competition matches (You can check the docs to see which version will be used)
As such it is possible that some behaviour may not match that of the official competition matches
```

controllers/competition_supervisor/competition_supervisor.py
WillB97: One small issue, what should the return type of the getter be since in our stubs we've created the _SimulationMode enum while in webots it appear to give the parameter directly to SWIG which is an enum in C.

PeterJCLaw: Hrm. I wouldn't want to make `_SimulationMode` public as that would imply that it could be used directly (which it obviously can't be).

WillB97: Also when running in webots the type would be `int` and `_SimulationMode` would be undefined

PeterJCLaw: I think the solution here is probably to use a `NewType`. It's not ideal, as we'll need to import the type definition with a type check guard, but it preserves the rest of the safety we currently get from the enum.

PeterJCLaw: I've put together a fix in https://github.com/srobo/competition-simulator/pull/262

PeterJCLaw: 5ba0d8d00ac13a395d7b6a2ae60c804220d7d309

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -134,6 +137,26 @@ def remove_unused_robots(supervisor: Supervisor) -> None:
         robot.remove()
 
 
+def get_simulation_run_mode(supervisor: Supervisor) -> 'SimulationMode':
+    if supervisor.getDevice("2021a-compatibility") is None:
```
PeterJCLaw, line 141: I'm a little surprised that this doesn't emit a `TypeError` under `2020b-rev1`, yet it seems that it doesn't. Presumably we're saved by the Supervisor robot not having any devices?
That might be worth a comment, especially as the error (if we did start seeing it) wouldn't be obvious to diagnose.

WillB97: Yes strangely the `TypeError` only seems to occur when the return value is assigned to a variable.
Perhaps something like:
```suggestion
    # webots 2020b is buggy and can raise TypeError when passed a str,
    # however this isn't occuring when the returned value isn't assigned to a variable
    if supervisor.getDevice("2021a-compatibility") is None:
```

PeterJCLaw: I seriously doubt that whether or not this is assigned to a variable is going to be affecting whether or not a `TypeError` is thrown. If it is, then something _very_ odd is going on with your Python interpreter. I strongly suspect that there may have been confounding effects when you observed that behaviour.

I've just tested assigning this to a local variable (and then using that in the conditional) and it still works fine.

Unless we know exactly why we're not seeing that error here (as noted I suspect it's because this `Robot` doesn't actually have any devices) I would encourage that we not commit to an explanation.

WillB97: Yeah I can't currently replicate the issue so we should probably just stick to the first line comment

WillB97: see 5cf09f549c0d9819588fe03836b169200903049d

controllers/competition_supervisor/competition_supervisor.py
WillB97: see 850b6e3f9a031f0d7187c490e3837875c9d8b00d

RealOrangeOne (approved):

PR #254, opened 2020-12-19
Add recording controls

PeterJCLaw:
This introduces a mechanism to control the duration of a match as well as the configuration of the recording from the match file. This is partly an experiment to see what it feels like to put more of this data into the match file, but also a way to make it easier for us to configure the recording when developing the simulator.

~If we like this approach it might be reasonable to also put the output paths for the recordings into this data structure too.~ Ended up deciding against this to avoid putting host specific data into files which are going to end up in the compstate.

This is part 2 of 3 towards match automation:
- Record match & scoring data (#244)
- Add recording controls (this PR #254)
- Automate running of individual matches (#256)


WillB97:
While I don't see much need to configure video quality since there is almost no case for using anything other than 100, I think placing the output paths for both recordings and logs into the match file is an important step to achieving a universal configuration file.

Also I've had issues with attempting to record at a resolution greater than the screen resolution.

PeterJCLaw:
> While I don't see much need to configure video quality since there is almost no case for using anything other than 100,

Agree. I'd initially thought that the quality parameter being too high meant that I was hitting a limitation of my hardware due to getting errors like the following when making recordings:
```
[libx264 @ 0x562cf1ba9840] Error: 2pass curve failed to converge
[libx264 @ 0x562cf1ba9840] target: 20250.00 kbit/s, expected: 3339.51 kbit/s, avg QP: 0.0252
[libx264 @ 0x562cf1ba9840] try reducing target bitrate
```
I've since discovered that these seem to be harmless (see 1e97dea).

> I think placing the output paths for both recordings and logs into the match file is an important step to achieving a universal configuration file.

So I'm currently _not_ planning to put the logs paths in there, mostly as I don't see that as being needed. Instead it seems reasonable that the runner script references their location and moves them onwards. I'm probably going to do something similar for the recordings, likely making the recording path relative to the `ARENA_ROOT` (rather than `REPO_ROOT`) along the way.

> Also I've had issues with attempting to record at a resolution greater than the screen resolution.

Could you elaborate on what those issues are? I've been recording fine at 1920×1080 despite that being above my screen res (at least in one dimension).

WillB97:
> Agree. I'd initially thought that the quality parameter being too high meant that I was hitting a limitation of my hardware due to getting errors like the following when making recordings:
> 
> ```
> [libx264 @ 0x562cf1ba9840] Error: 2pass curve failed to converge
> [libx264 @ 0x562cf1ba9840] target: 20250.00 kbit/s, expected: 3339.51 kbit/s, avg QP: 0.0252
> [libx264 @ 0x562cf1ba9840] try reducing target bitrate
> ```
> 
> I've since discovered that these seem to be harmless (see [1e97dea](https://github.com/srobo/competition-simulator/commit/1e97dea62a0615feadc1f6505af8844b75933d6b)).

Indeed, these errors it because webots' bitrate calculation in requesting an extraordinarily high bitrate of 20Mbit/s which would have made each match a 300MB recording.

> > Also I've had issues with attempting to record at a resolution greater than the screen resolution.
> 
> Could you elaborate on what those issues are? I've been recording fine at 1920×1080 despite that being above my screen res (at least in one dimension).

On closer inspection my issues were just visual glitches when webots resizes the window and the recording comes out fine.

WillB97 (approved):

PR #256, opened 2020-12-19
Automate running of individual matches

### LINKED ISSUE #189: Automate collation of logs by team

### LINKED ISSUE #210: Automate running of matches

PeterJCLaw:
This (nearly) achieves full automation for single match runs.

I've gone a different route than I'd initially envisaged in #210, mostly because I'd overlooked the issue of symlinks not working on Windows (which I'm assuming we still want to support as a host platform).

This is based on `recording-controls` (#254) mostly to avoid conflicts; the actual changes this depends on are (I think) all in (#244). This unfortunately will definitely conflict with #255, but we can deal with that.

~One thing I'm not keen on here is that to test the full setup at all you need robot code which is in an archive, yet we don't have a good way to take an existing directory and archive it correctly. That might be handy for development.~

This is part 3 of 3 towards match automation:
- Record match & scoring data (#244)
- Add recording controls (#254)
- Automate running of individual matches (this PR #256)

Fixes #189.
Fixes #210.


script/run-comp-match
```diff
@@ -16,20 +17,85 @@ sys.path.insert(1, str(REPO_ROOT / 'modules'))
...
+    for zone_id, tla in enumerate(match_data.teams):
+        if tla is None:
+            # no team in this zone
+            continue
+
+        log_filname = controller_utils.get_robot_log_filename(zone_id)
+        log_path = get_zone_path(zone_id) / log_filname
+
+        team_dir = archives_dir / tla
+        team_dir.mkdir()
```
WillB97, line 86:
```suggestion
        team_dir.mkdir(exist_ok=True)
```
Since after the first match for each team this folder will exist and we aren't catching the FileExistsError

PeterJCLaw: Huh, for some reason I'd thought that this was the default. Seems not.

script/run-comp-match
```diff
@@ -16,20 +17,85 @@ sys.path.insert(1, str(REPO_ROOT / 'modules'))
 import controller_utils  # noqa:E402 # isort:skip
 from controller_utils import NUM_ZONES  # isort:skip
 
+NUM_CONTROLLERS = NUM_ZONES + 2
```
WillB97, line 20: Is there a purpose for this line as the constant appears to be unused.

WillB97:
> * [ ]  work out where the recordings should end up (and change that)

According to #210 recordings end up in archives_dir / "recordings", if we are going to export ARENA_ROOT to point at the archives_dir then get_recording_path needs to be modified to remove the date component during automated matches (likely when match_num is set).

> One thing I'm not keen on here is that to test the full setup at all you need robot code which is in an archive, yet we don't have a good way to take an existing directory and archive it correctly. That might be handy for development.

Don't we have a script for the competitiors to do this?

PeterJCLaw:
> > * [ ]   work out where the recordings should end up (and change that)
> 
> According to #210 recordings end up in archives_dir / "recordings", if we are going to export ARENA_ROOT to point at the archives_dir then get_recording_path needs to be modified to remove the date component during automated matches (likely when match_num is set).

`get_recording_path` will need to be modified, though I don't expect that `ARENA_ROOT` will point at `archives_dir`. Doing that would mix our reference source for the team archives with the working directory for a match, which feels like a bad direction to head in. Rather I plan that `ARENA_ROOT` will likely be a temporary directory per match. This ensures it's clean at the start and also trivially allows for matches to be run in parallel if we wanted to.

> > One thing I'm not keen on here is that to test the full setup at all you need robot code which is in an archive, yet we don't have a good way to take an existing directory and archive it correctly. That might be handy for development.
> 
> Don't we have a script for the competitiors to do this?

Are you thinking of https://github.com/srobo/brain-zip-packager/? That doesn't work for the simulator.

Likely all this needs to be is a script containing `zip robot.zip *.py`, but I can never remember that without looking it up and/or get the folder hierarchy handling wrong. I expect competitors will be doing this by hand, so this is just for me being lazy in development.

script/run-comp-match
PeterJCLaw: Nope. 11dc5e6

modules/controller_utils/__init__.py
```diff
@@ -11,9 +11,6 @@
 # Root directory of the specification of the Arena (and match)
 ARENA_ROOT = Path(os.environ.get('ARENA_ROOT', REPO_ROOT.parent))
```
WillB97, line 12: While this wasn't added here we are now recommending its use, setting ARENA_ROOT as an environment variable only works when setting an absolute path. If a relative path is used every file takes that path relative to itself

WillB97: This change probably should be its own PR

WillB97: See #266

WillB97:
Currently running `run-comp-match` with one of the robots replaced with a dash throws an error in competition_supervisor
```
Using '/tmp/tmp0yfut3k3match-24' as the arena
INFO: sr_controller: Starting controller: /home/will/work/notes/SR_simulator/competition-simulator/venv/bin/python -u sr_controller.py
INFO: sr_controller: Starting controller: /home/will/work/notes/SR_simulator/competition-simulator/venv/bin/python -u sr_controller.py
INFO: competition_supervisor: Starting controller: /home/will/work/notes/SR_simulator/competition-simulator/venv/bin/python -u competition_supervisor.py
INFO: territory_controller: Starting controller: /home/will/work/notes/SR_simulator/competition-simulator/venv/bin/python -u territory_controller.py
No robot controller found for zone 0
WARNING: 'sr_controller' controller exited with status: 1.
INFO: Start HTML5 animation export

INFO: You are entering fullscreen mode.
INFO: sr_controller: Terminating.
INFO: competition_supervisor: Terminating.
INFO: territory_controller: Terminating.

Waiting for 0
Zone 0 ready
Zone 1 ready
Saving animation to /tmp/tmp0yfut3k3match-24/recordings/match-24.html
Saving video to /tmp/tmp0yfut3k3match-24/recordings/match-24.mp4
===========
Match start
===========
Failed to get Webots node for zone 0 (id: 5)
Traceback (most recent call last):
  File "competition_supervisor.py", line 187, in <module>
    main()
  File "competition_supervisor.py", line 183, in main
    run_match(supervisor)
  File "competition_supervisor.py", line 143, in run_match
    for _, robot in get_robots(supervisor):
  File "competition_supervisor.py", line 102, in get_robots
    raise ValueError(msg)
ValueError: Failed to get Webots node for zone 0 (id: 5)

Simulation errored (exit code 1). Competition supervisor logs are above.
```

modules/controller_utils/__init__.py
PeterJCLaw: Ah, that's a shame. Ultimately absolute paths are what is needed here, so I'm not inclined to change this. We could add a warning if we detect the user has passed a relative path though.

PeterJCLaw:
> Currently running `run-comp-match` with one of the robots replaced with a dash throws an error in competition_supervisor

That's weird -- nothing about that should have changed here. Will have a look.

PeterJCLaw:
> > Currently running `run-comp-match` with one of the robots replaced with a dash throws an error in competition_supervisor
> 
> That's weird -- nothing about that should have changed here. Will have a look.

From looking in the history and at what's causing this, I think that this was caused by #205. The underlying issue is that we try to use `get_robots` after we've removed the dead nodes. Either that function needs to allow for that or we need to rework the way we use the data. I'll pick that up separately.

README.md
```diff
@@ -81,39 +81,28 @@ In order to run competition matches you'll need to:
...
-3. To put the competitors' code into the right places for each of the corner zones, run
+3. Run the match:
 
     ```bash
-    competition-simulator/script/prepare-comp-match <directory containing team code> <match number> <Zone 0 TLA> <Zone 1 TLA> <Zone 2 TLA> <Zone 3 TLA>
+    competition-simulator/script/run-comp-match <directory containing team code> <match number> <Zone 0 TLA> <Zone 1 TLA> <Zone 2 TLA> <Zone 3 TLA>
     ```
 
-    Using a dash instead of a TLA if a robot is not present. This provides filenames and sets up the directory structure:
+    Using a dash instead of a TLA if a robot is not present.
```
RealOrangeOne, line 90:
```suggestion
    Note: Use a dash instead of a TLA if a robot is not present.
```

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -63,6 +52,17 @@ def record_video(supervisor: Supervisor, file_path: Path) -> Iterator[None]:
         print("Movie failed to record")
 
 
+@contextlib.contextmanager
+def propagate_exit_code(supervisor: Supervisor) -> Iterator[None]:
```
RealOrangeOne, line 56: Worth a docstring here?

script/run-comp-match
```diff
@@ -0,0 +1,200 @@
...
+        controller_utils.ARENA_ROOT = Path(tmpdir_name)
+        try:
+            yield
+        finally:
+            os.environ['ARENA_ROOT'] = str(arena_root)
+            controller_utils.ARENA_ROOT = arena_root
+
+
+def prepare_match(archives_dir: Path, match_data: controller_utils.MatchData) -> None:
+    controller_utils.get_mode_file().write_text('comp\n')
```
RealOrangeOne, line 58: Given the data is stripped, we can probably remove the newline here

```suggestion
    controller_utils.get_mode_file().write_text('comp')
```

script/run-comp-match
```diff
@@ -0,0 +1,200 @@
...
+        BOLD = '\033[1m'
+        FAIL = '\033[91m'
+        ENDC = '\033[0m'
+
+        log_text = controller_utils.get_competition_supervisor_log_filepath().read_text()
+
+        print("\n" + FAIL + log_text + ENDC)  # noqa:T001
+        print(  # noqa:T001
+            f"{BOLD}{FAIL}Simulation errored (exit code {e.returncode}). "
+            f"Competition supervisor logs are above.{ENDC}",
```
RealOrangeOne, line 96: Rather than saying the logs are above, worth putting the logs below?

script/run-comp-match
```diff
@@ -0,0 +1,200 @@
...
+    recordings_dir = archives_dir / 'recordings'
+    recordings_dir.mkdir(exist_ok=True)
+
+    recording_stem = controller_utils.get_recording_stem()
+
+    for path in recording_stem.parent.glob(f'{recording_stem.name}.*'):
+        shutil.copy(path, recordings_dir)
+
+    try:
+        shutil.copytree(recording_stem.parent / 'textures', recordings_dir / 'textures')
```
RealOrangeOne, line 151: Worth a comment here about why we're copying textures?

script/run-comp-match
```diff
@@ -0,0 +1,200 @@
...
+        args.match_num,
+        tlas,
+        args.duration,
+        recording_config=None,  # use the default
+    )
+
+
+@contextlib.contextmanager
+def temporary_arena_root(suffix: str) -> Iterator[None]:
+    arena_root = controller_utils.ARENA_ROOT
```
RealOrangeOne, line 45: Might be nicer to name this `previous_arena_root`, mostly so subsequent usages don't seem strange.

RealOrangeOne (commented):
This looks so much easier to do compared to last year!

Just a couple small comments / queries.

script/run-comp-match
WillB97, line 45: Is there a reason we are backing up and restoring `ARENA_ROOT` given that we never use the old value

script/run-comp-match
```diff
@@ -0,0 +1,200 @@
...
+
+def run_match() -> None:
+    try:
+        subprocess.check_call([
+            'webots',
+            '--batch',
+            '--stdout',
+            '--stderr',
+            '--minimize',
+            str(REPO_ROOT / 'worlds' / 'Arena.wbt'),
```
WillB97, line 84: Do we want to explicitly set the simulation mode, after the first time-step the competition simulator takes over but this gets it started.
```suggestion
            '--mode=realtime',
            str(REPO_ROOT / 'worlds' / 'Arena.wbt'),
```

controllers/competition_supervisor/competition_supervisor.py
PeterJCLaw: f226eb4

script/run-comp-match
PeterJCLaw: This isn't strictly new -- it's part of the code which was moved from `script/prepare-comp-match`. I'm also tempted to leave this as it is since having a trailing newline generally makes text files easier to work with.

script/run-comp-match
```diff
@@ -0,0 +1,200 @@
...
+    parser.add_argument(
+        'archives_dir',
+        help=(
+            "The directory containing the teams' robot code, as Zip archives "
+            "named for the teams' TLAs. This directory will also be used as the "
+            "root for storing the resulting logs and recordings."
+        ),
+        type=Path,
+    )
+    parser.add_argument('match_num', help="The number of the match to run.")
```
WillB97, line 167: We don't appear to be validating that the match number is an int anywhere.
```suggestion
    parser.add_argument(
        'match_num',
        help="The number of the match to run.",
        type=int,
    )
```

script/run-comp-match
PeterJCLaw: Does a568c1a help explain why they're in the order they are?

script/run-comp-match
PeterJCLaw: 2a83607

script/run-comp-match
PeterJCLaw: 3cc08ee

README.md
PeterJCLaw: As this part isn't new in this PR I'm going to leave this. I agree that that would be a better sentence though, so feel free to push it on `master` (ideally after this PR merges to avoid conflicts).

script/run-comp-match
PeterJCLaw: Mostly so that this context manager is easily re-usable in the event that it gets copied/moved for re-use.

script/run-comp-match
PeterJCLaw: Is this going to cause issues if the name changes, as we saw with 2021a?

script/run-comp-match
PeterJCLaw: Good spot :) 0259858

script/run-comp-match
WillB97: While technically yes, the change in 2021a was because of enable rendering was moved to a separate variable so it's unlikely the `realtime` would be renamed.

WillB97 (approved):

script/run-comp-match
RealOrangeOne: Works for me!

RealOrangeOne (approved):

PR #260, opened 2020-12-23
Fix distance sensor returning max distance incorrectly

WillB97:
Certain surfaces of the arena were not being detected by the distance sensors resulting in the sensors returning their maximum distance even when next to, and parallel to, the wall.
To remedy this the arena has been reconstructed using simple block shapes. It was additionally discovered that webots currently suffers this same issue with the TexturedBox object.


PeterJCLaw:
Not sure I'm really the right person to review this as I made most of the changes?

sedders123 (dismissed):
Changes seem reasonable and I've had a play locally and the numbers reported by the sensors seem to be correct 🎉

sedders123:
This image doesn't look right to me, back left doesn't look like it should be 2
image

PeterJCLaw:
> This image doesn't look right to me, back left doesn't look like it should be 2
> image

Bear in mind that at high angles to the walls we don't expect the sensors to work. Can you reproduce with the sensor more perpendicular to the wall?

sedders123:
> Bear in mind that at high angles to the walls we don't expect the sensors to work.
Ahh wasn't aware of that. When more perpendicular the numbers seen fine!

sedders123 (approved):

PR #261, opened 2020-12-23
Add various development notes

PeterJCLaw:
This aims to provide more detail for someone looking to develop the automation systems as well as the simulator in general.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,48 @@ pip install -r script/typing/requirements.txt
 
 You can then run all linting/type checking/tests in one go using `script/check`.
 
+### Running Webots
+
+While the default location that our controllers look for `robot.py` files is the
+directory above the repo, that is not particularly convenient for development.
+Instead you may wish to run Webots having set the `ARENA_ROOT` environment
+variable to a suitable location.
+
+For example, you may find it convenient to have a `robots` directory within the
+repo and then have a number of code and arena directories within that:
+
+```
+robots
+├── arena  # A development arena with a single (symlinked) robot
+│   └── robot.py -> ../ultrasounds/robot.py
+├── brakes-arena  # A competition style arena with two robots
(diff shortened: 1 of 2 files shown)


WillB97:
Do we want to add a note about how to create robot archives as well?

RealOrangeOne (approved):

PeterJCLaw:
> Do we want to add a note about how to create robot archives as well?

Done in ee06104e6683404783dfcc57100e84265aa32e59. In hindsight I'm not sure that's actually that useful.

I think the next step here is probably to re-evaluate the README as a whole and re-structure it as more general documentation. There's a lot of things I've ended up covering here from a development standpoint which should actually just be plain documented. I think this is better than nothing for now though.

WillB97 (approved):

WillB97:
We should probably mention that the ARENA_ROOT environment variable cannot be passed to `run-comp-match`

README.md
```diff
@@ -43,6 +43,82 @@ pip install -r script/typing/requirements.txt
...
+└── ultrasounds  # Some robot code for testing ultrasound sensors
+    └── robot.py
+```
+
+Given a setup like the above, running Webots such that it picks up on one of the
+arena directories is possible through setting the `ARENA_ROOT` when launching
+the webots process:
+
+```
+$ ARENA_ROOT=robots/brakes-arena webots --mode=pause worlds/Arena.wbt
```
PeterJCLaw, line 78: Need to clarify here that the given path must be an absolute path.

PeterJCLaw: a78aec9

PeterJCLaw:
> We should probably mention that the ARENA_ROOT environment variable cannot be passed to `run-comp-match`

done in a78aec9

WillB97 (approved):

PR #262, opened 2020-12-23
Make it possible to import SimulationMode for type checking

PeterJCLaw:
This still needs to be guarded by `TYPE_CHECKING` as it's not really something which exists in Webots, however it is useful to be able to have this for use in used-defined signatures.


WillB97 (approved):
That should solve the problem

PR #263, opened 2020-12-27
Reduce logging IO

WillB97:
The match data does not need to be written to disk every timestep.
This branch now only writes the claim log after a change has occurred.

Writing the claim log to disk every timestep is 15,000 writes in a two-minute match.
By only writing after a territory claim has occurred this will be reduced to a handful of writes in most cases.

On my setup, this reduced the time to run a two-minute match by ~25% (bringing it inline with a match run on master) since the file was stored on a HDD.

### DIFF
--- a/controllers/territory_controller/territory_controller.py
+++ b/controllers/territory_controller/territory_controller.py
@@ -58,8 +58,8 @@ class ClaimLog:
         }
 
         self._log: List[Tuple[StationCode, Claimant, float]] = []
-        # use < 0 so that the log format is written at least once
-        self._saved_log_length: int = -1
+        # starting with a dirty log ensure the structure is written for every match
+        self._log_is_dirty = True
 
     def get_claimant(self, station_code: StationCode) -> Claimant:
         return self._station_statuses[station_code]
@@ -71,6 +71,7 @@ class ClaimLog:
         claim_time: float,
     ) -> None:
         self._log.append((station_code, claimed_by, claim_time))
+        self._log_is_dirty = True
         print(f"{station_code} CLAIMED BY {claimed_by} AT {claim_time}s")  # noqa:T001
         self._station_statuses[station_code] = claimed_by
 
@@ -78,7 +79,7 @@ class ClaimLog:
         if not self._record_arena_actions:
             return
 
-        if self._saved_log_length == len(self._log):
+        if self._log_is_dirty is True:
             # don't write the log if nothing new has happened
             return
 
@@ -91,7 +92,7 @@ class ClaimLog:
             for station_code, claimed_by, claim_time in self._log
         ]})
 
-        self._saved_log_length = len(self._log)
+        self._log_is_dirty = False
 
 
 class TerritoryController:
--- a/controllers/territory_controller/territory_controller.py
+++ b/controllers/territory_controller/territory_controller.py
@@ -58,7 +58,7 @@ class ClaimLog:
         }
 
         self._log: List[Tuple[StationCode, Claimant, float]] = []
-        # starting with a dirty log ensures the structure is written for every match
+        # Starting with a dirty log ensures the structure is written for every match.
         self._log_is_dirty = True
 
     def get_claimant(self, station_code: StationCode) -> Claimant:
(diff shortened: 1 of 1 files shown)


RealOrangeOne (commented):
Nice catch!

I wonder if this would be simpler by just using an `is_dirty` flag to store whether writes need to happen, and just set that on write to file / log?

WillB97:
Since the log is only appended through the ClaimLog.log_territory_claim() method an is_dirty flag could be set in there.

RealOrangeOne:
Yeah that code looks about like what I was expecting.

Although it does raise another thought: Should we just save the data to the file when logging? That way they can't get out of sync, and there's no need to store things in memory even. It would make claims _marginally_ slower and more expensive, but not enough to matter. Likely also reduces a lot of complexity.

PeterJCLaw:
> Should we just save the data to the file when logging? That way they can't get out of sync, and there's no need to store things in memory even. It would make claims _marginally_ slower and more expensive, but not enough to matter. Likely also reduces a lot of complexity.

Maaybe, though until we've solved the questions around what happens if two teams claim a zone during the same timestep I don't think we should consider a single timestep as an atomic unit.

> no need to store things in memory even.

I had anticipated that we might want the simulator to do a realitime overlay of the scores or something at some point, which was part of why this is done the way it is.
Doing a full write avoids needing read back in the scores each time (along with questions around what happens if that fails).

controllers/territory_controller/territory_controller.py
```diff
@@ -69,13 +71,18 @@ def log_territory_claim(
...
         self._log.append((station_code, claimed_by, claim_time))
+        self._log_is_dirty = True
         print(f"{station_code} CLAIMED BY {claimed_by} AT {claim_time}s")  # noqa:T001
         self._station_statuses[station_code] = claimed_by
 
     def record_captures(self) -> None:
         if not self._record_arena_actions:
             return
 
+        if self._log_is_dirty is True:
```
RealOrangeOne, line 82:
```suggestion
        if self._log_is_dirty:
```

controllers/territory_controller/territory_controller.py
```diff
@@ -69,13 +71,18 @@ def log_territory_claim(
...
         self._log.append((station_code, claimed_by, claim_time))
+        self._log_is_dirty = True
         print(f"{station_code} CLAIMED BY {claimed_by} AT {claim_time}s")  # noqa:T001
         self._station_statuses[station_code] = claimed_by
 
     def record_captures(self) -> None:
         if not self._record_arena_actions:
             return
 
+        if self._log_is_dirty:
```
sedders123, line 82: Should this check be inverted? The log is dirty so we want to write it to disk rather than exiting out early.
```suggestion
        if not self._log_is_dirty:
```

WillB97: Very true

controllers/territory_controller/territory_controller.py
```diff
@@ -58,6 +58,8 @@ def __init__(self, record_arena_actions: bool) -> None:
         }
 
         self._log: List[Tuple[StationCode, Claimant, float]] = []
+        # starting with a dirty log ensures the structure is written for every match
```
PeterJCLaw, line 61: Really minor - would be great to have comments as proper sentences:
```suggestion
        # Starting with a dirty log ensures the structure is written for every match.
```

controllers/territory_controller/territory_controller.py
```diff
@@ -69,13 +71,18 @@ def log_territory_claim(
...
+        self._log_is_dirty = True
         print(f"{station_code} CLAIMED BY {claimed_by} AT {claim_time}s")  # noqa:T001
         self._station_statuses[station_code] = claimed_by
 
     def record_captures(self) -> None:
         if not self._record_arena_actions:
             return
 
+        if not self._log_is_dirty:
+            # don't write the log if nothing new has happened
```
PeterJCLaw, line 83:
```suggestion
            # Don't write the log if nothing new has happened.
```

PeterJCLaw (approved):

sedders123 (approved):

PR #264, opened 2020-12-29
Allow infinite LED changes within the same timestep

WillB97:
Since all changes within a timestep visually apply together changing multiple LEDs within a single timestep is not an epilepsy issue so we should be limiting it.

This also has the side effect of allowing LEDs to be set at the start of the match (although not during the first half second as webots time starts at 0).

### DIFF
--- a/modules/sr/robot/output_frequency_limiter.py
+++ b/modules/sr/robot/output_frequency_limiter.py
@@ -15,7 +15,9 @@ class OutputFrequencyLimiter:
         now = self._webot.getTime()
         diff = now - self._last_change
 
-        # allow infinite changes within a timestep
+        # Changes within a timestep always all happen during the same
+        # render cycle and therefore cannot contribute to additional strobing.
+        # It's therefore safe to allow several changes within the same timestep.
         if diff == 0:
             return True
 
--- a/modules/sr/robot/output_frequency_limiter.py
+++ b/modules/sr/robot/output_frequency_limiter.py
@@ -14,6 +14,11 @@ class OutputFrequencyLimiter:
     def can_change(self) -> bool:
         now = self._webot.getTime()
         diff = now - self._last_change
+        print(diff)
+
+        # allow infinite changes within a timestep
+        if diff == 0:
+            return True
 
         if diff < MIN_TIME_BETWEEN_CHANGES:
             return False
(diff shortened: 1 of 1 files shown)


RealOrangeOne:
Given that it's the main intention of the PR, I wonder if we should be checking `now == 0` rather than `diff`. It's possible allowing setting inside the same timestep eventually leads to weirdness.

WillB97:
> Given that it's the main intention of the PR, I wonder if we should be checking `now == 0` rather than `diff`.
 
`now == 0` only occurs at `time == 0` which only fixes setting LEDs at the start of the match, not changing LEDs at the same time.

> It's possible allowing setting inside the same timestep eventually leads to weirdness.

Given the LEDs are independent devices setting different LEDs will be fine. Setting the same LED multiple times appears to just take the final value.

modules/sr/robot/output_frequency_limiter.py
```diff
@@ -15,6 +15,10 @@ def can_change(self) -> bool:
         now = self._webot.getTime()
         diff = now - self._last_change
 
+        # allow infinite changes within a timestep
```
PeterJCLaw, line 18: I think it would be useful to explain why this is ok. Assuming my understanding is correct, something like this might be more useful to future readers:
```suggestion
        # Changes within a timestep always all happen during the same
        # render cycle and therefore cannot contribute to additional strobing.
        # It's therefore safe to allow several changes within the same timestep.
```

PeterJCLaw (approved):

modules/sr/robot/output_frequency_limiter.py
WillB97: That is much better wording

PeterJCLaw (approved):

modules/sr/robot/output_frequency_limiter.py
```diff
@@ -15,6 +15,12 @@ def can_change(self) -> bool:
         now = self._webot.getTime()
         diff = now - self._last_change
 
+        # Changes within a timestep always all happen during the same
```
sedders123, line 18:
```suggestion
        # Changes within a timestep will happen during the same
```

sedders123 (approved):
Looks good! Just one bit of the comment that sounded a little odd to me but take or leave the change :)

PR #265, opened 2020-12-29
Change rate limiter text to refer to led pin

WillB97:
Error messages are now in the form: `requested setting LED on pin X to False` instead of using their device names.

### DIFF
--- a/modules/sr/robot/ruggeduino.py
+++ b/modules/sr/robot/ruggeduino.py
@@ -35,12 +35,7 @@ def init_ruggeduino_array(webot: Robot) -> 'List[Ruggeduino]':
 
     limiter = OutputFrequencyLimiter(webot)
     digital_output_array = [
-        Led(
-            webot,
-            name,
-            limiter,
-            pin,
-        )
+        Led(webot, name, limiter, pin)
         for pin, name in enumerate(
             led_names,
             start=Ruggeduino.DIGITAL_PIN_START + len(digital_input_array),
--- a/modules/sr/robot/ruggeduino.py
+++ b/modules/sr/robot/ruggeduino.py
@@ -39,9 +39,12 @@ def init_ruggeduino_array(webot: Robot) -> 'List[Ruggeduino]':
             webot,
             name,
             limiter,
-            index + Ruggeduino.DIGITAL_PIN_START + len(digital_input_array),
+            pin,
         )
-        for index, name in enumerate(led_names)
+        for pin, name in enumerate(
+            led_names,
+            start=Ruggeduino.DIGITAL_PIN_START + len(digital_input_array),
+        )
     ]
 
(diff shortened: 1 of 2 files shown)


modules/sr/robot/ruggeduino_devices.py
```diff
@@ -72,16 +72,18 @@ def __init__(
         webot: Robot,
         device_name: str,
         limiter: OutputFrequencyLimiter,
+        pin_num: int,
     ) -> None:
         self._name = device_name
```
PeterJCLaw, line 77: I think `_name` is no longer used -- could we drop it?

WillB97: Indeed `device_name` is only used to get the sensor object at initialisation so there is no need to store it.

WillB97: Corrected in 03ccbc09bd43e5a3818d31d6632f2a9961d7ee51

modules/sr/robot/ruggeduino.py
```diff
@@ -34,7 +34,15 @@ def init_ruggeduino_array(webot: Robot) -> 'List[Ruggeduino]':
...
     limiter = OutputFrequencyLimiter(webot)
-    digital_output_array = [Led(webot, name, limiter) for name in led_names]
+    digital_output_array = [
+        Led(
+            webot,
+            name,
+            limiter,
+            index + Ruggeduino.DIGITAL_PIN_START + len(digital_input_array),
+        )
+        for index, name in enumerate(led_names)
```
PeterJCLaw, line 44: Maybe pass `start` to `enumerate` to avoid this inline maths? Not sure which is clearer.

WillB97: That does make it clearer what the `index` actually is.

modules/sr/robot/ruggeduino.py
```diff
@@ -34,7 +34,18 @@ def init_ruggeduino_array(webot: Robot) -> 'List[Ruggeduino]':
...
 
     limiter = OutputFrequencyLimiter(webot)
-    digital_output_array = [Led(webot, name, limiter) for name in led_names]
+    digital_output_array = [
+        Led(
+            webot,
+            name,
+            limiter,
+            pin,
+        )
```
PeterJCLaw, line 43: Miiight be worth unwrapping this now?
```suggestion
        Led(webot, name, limiter, pin)
```

sedders123 (approved):

PeterJCLaw (approved):
:+1:

Tested with:
``` python
from sr.robot import Robot

R = Robot()

while True:
    R.ruggeduinos[0].digital_write(5, True)
    R.sleep(0.1)
    R.ruggeduinos[0].digital_write(5, False)
    R.sleep(0.1)
```

PR #266, opened 2020-12-29
Allow ARENA_ROOT to be a path relative to REPO_ROOT

WillB97:
In #261 we show setting `ARENA_ROOT` to a relative path, this currently does not work with the variable being taken relative to each file accessing it.

This changes the behaviour of the `ARENA_ROOT` environment variable to be either an absolute path or relative to `REPO_ROOT`.

### DIFF
--- a/modules/controller_utils/__init__.py
+++ b/modules/controller_utils/__init__.py
@@ -9,7 +9,7 @@ from pathlib import Path
 REPO_ROOT = Path(__file__).resolve().parent.parent.parent
 
 # Root directory of the specification of the Arena (and match)
-ARENA_ROOT = Path(os.environ.get('ARENA_ROOT', REPO_ROOT.parent))
+ARENA_ROOT = (REPO_ROOT / os.environ.get('ARENA_ROOT', '..')).resolve()
 
 
 ROBOT_IDS_TO_CORNERS = {


PeterJCLaw:
I don't think this is a good idea as it forces all `ARENA_ROOT` values to be relative to the repo. The original intent was that it should be unspecified. That allows for either absolute paths or paths relative to the current directory.
Perhaps the example should have included `$PWD`? However that's platform specific (I think would need `%CD%` on Windows) which was why it wasn't included initially.

WillB97:
Webots sets the current directory to each controller file so the directory of that webots was launched from is not available. This PR aims to make so using a relative path in `ARENA_ROOT` at least provides a consistent location.
The current behaviour makes relative paths totally unusable since this does not point to a universal location from all files attempting to access arena files.

PeterJCLaw:
> Webots sets the current directory to each controller file so the directory of that webots was launched from is not available. This PR aims to make so using a relative path in `ARENA_ROOT` at least provides a consistent location.
> The current behaviour makes relative paths totally unusable since this does not point to a universal location from all files attempting to access arena files.

Ah, I'd not realised that. I suggest then that we reject relative paths and change the docs to clarify that paths should be absolute.

WillB97:
Since #269 supersedes this we can close this.

PR #267, opened 2020-12-30
Add a mailmap for consistent names

PeterJCLaw:
If I've picked you for review, it's because you're a committer whose naming is altered by the mailmap -- please approve to let us know you're happy with the spellings in the file.

So that this doesn't just get stuck, I'll revisit this PR in a week and remove any which haven't responded.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,9 @@
+Antoine Petty <github@antoinepetty.co.uk> <36314396+antoinepetty@users.noreply.github.com>
+
+Phil Renouf <PhilRenouf54@gmail.com>
+Phil Renouf <PhilRenouf54@gmail.com> <PhilR54@users.noreply.github.com>
+
+Jake Howard <git@theorangeone.net> <RealOrangeOne@users.noreply.github.com>
+
+Will Barber <will@rudi.org.uk>
+Will Barber <will@rudi.org.uk> <WillB97@users.noreply.github.com>


WillB97 (approved):

.mailmap
```diff
@@ -0,0 +1,11 @@
+Antoine Petty <a.petty@ncl.ac.uk> <36314396+antoinepetty@users.noreply.github.com>
```
antoinepetty, line 1:
```suggestion
Antoine Petty <github@antoinepetty.co.uk>
Antoine Petty <github@antoinepetty.co.uk> <36314396+antoinepetty@users.noreply.github.com>
```

antoinepetty (changes requested):

.mailmap
antoinepetty: Is this to attribute people to the repo? I don't really understand

PeterJCLaw: Currently if you run `git log` in this repo you get lots of different names for what are actually the same people. This is most evident when running `git shortlog --summary --email`.

Adding this `.mailmap` file tells git which of those identities are in fact the same person as well as what true name to use for them.

PeterJCLaw: The `@users.noreply.github.com` addresses come from when you use GitHub's UI to make changes. You can fix that by setting which of your email addresses GitHub should use for those commits.

antoinepetty: Ah so my suggested changes are actually very unhelpful??

PeterJCLaw: Sorry, no, they were useful! I wanted to force-push to remove the original commit so that your other email address wasn't part of the history here. I'd copied that from the docs repo as I'd assumed it was your preferred one.

antoinepetty (approved):

PhilR54 (approved):

RealOrangeOne (approved):
Looks fine to me.

Although not sure if the duplication of people is intentional?

PeterJCLaw:
> Although not sure if the duplication of people is intentional?

Do you mean that there's more than one entry for some people? Note that a line such as `Name <email1> <email2>` won't change the display name for other occurrences of `<email1>`, so that needs to be listed separately.

PR #268, opened 2020-12-30
Reduce logging IO when saving match history

PeterJCLaw:
This is a replica of #263, but applied to #244 instead. This PR exists as a marker only.


Heath123:
This showed on the globe of the GuitHub main page

PR #270, opened 2020-12-30
Skip missing robot nodes when sending the start signal

PeterJCLaw:
Since we remove robot nodes as part of the pre-match setup, it's expected by this point that there may be some missing nodes.

Fixes the issue highlighted at https://github.com/srobo/competition-simulator/pull/256#issuecomment-752128650.


WillB97 (approved):
Solves the observed issue.

PR #271, opened 2020-12-31
Run a comp mode match in CI

PeterJCLaw:
This provides a very basic full-stack validation of both the competition supervisor and the main Robot API.

For the latter, we check that the robot logs are produced and contain a unique marker which the CI inserts. This ensures that the robot has run to completion without error.

We don't explicitly validate the comp match running, however it seems likely that errors would cause it to error-exit, which we would see here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,9 +124,14 @@ jobs:
       - run:
           name: Install dependencies
           command: |
+            # Use deadsnakes to get Python 3.7
+            echo deb http://ppa.launchpad.net/deadsnakes/ppa/ubuntu focal main > /etc/apt/sources.list.d/deadsnakes.list
+            echo deb-src http://ppa.launchpad.net/deadsnakes/ppa/ubuntu focal main >> /etc/apt/sources.list.d/deadsnakes.list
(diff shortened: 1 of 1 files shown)


RealOrangeOne:
The camera overlay is visible in the output, but isn't in recordings outside CI. That's slightly concerning?

.circleci/config.yml
```diff
@@ -113,6 +145,57 @@ jobs:
...
+            export DEBIAN_FRONTEND=noninteractive
+            apt-get update
+            apt-get install --yes zip python3.7{,-dev,-venv}
+
+      - run:
+          name: Prepare team code archive
+          command: |
+            mkdir archives
+            cd archives
+            sed -r 's/# sleep(.*)/print("sleep\1")/' ../controllers/example_controller/example_controller.py > robot.py
```
RealOrangeOne, line 171: What's this line for?

WillB97: That converts the comments about `sleep for X seconds` into print statements so that there is some output to gauge if the simulator is running or frozen.

PeterJCLaw: 4ff1d74

WillB97 (approved):
Although we should keep an eye on how this affects our credit usage on CircleCI given that this will increase our compute minutes.

PR #277, opened 2021-01-10
Convert interior walls to IndexedFaceSet to fix sonar

WillB97:
This is the same issue as #260 as `TexturedParallelepiped` objects are also affected.
There also appears to be a partial issue that affects `IndexedFaceSet` when the shape has faces which do not have at least on common vertex with all others.

As such each wall is raplaced by a pair of `IndexedFaceSet` containing only 2 faces each.

### DIFF
--- a/worlds/Arena.wbt
+++ b/worlds/Arena.wbt
@@ -280,43 +280,149 @@ Solid {
   name "South Wall"
 }
 Solid {
-  translation -2.9 0.15 -0.333
-  rotation 0.5773499358549823 -0.5773489358550934 0.5773519358547601 2.09439
+  # An issue with the DistanceSensor in webots-2020b-r1 
+  # causes certain faces to not be detected
+  # It appears that the issue does not occur in IndexedFaceSet
+  # when all faces have a common vertex with all others
   children [
-    DEF SHAPE Shape {
-      appearance PBRAppearance {
-        baseColor 0.095395215 0.22841774 0.8000001
-        roughness 1
-        metalness 0
-      }
-      geometry TexturedParallelepiped {
-        size 0.28 4.89 0.3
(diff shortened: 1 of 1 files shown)


RealOrangeOne (commented):
These walls definitely look visually the same.

How reliably does this fix the issue, and how testable is it?

WillB97:
Prior to this fix, you could drag the robot parallel to the wall and one end of each face would not be detected.
Repeating this across all faces with the fix, I found no location which wasn't detected.

Techically to fully test this we would need to test the full length of all walls at various angles and distances. But given that testing before and after the fix shows an improvement in results we can assume that this specific issue has been solved. Additionally since all the outer walls have already been replace for this issue the only other location would be the arena triangle, which performing the same test as above showed no errors.

worlds/Arena.wbt
```diff
@@ -280,43 +280,149 @@ Solid {
...
-      }
+    Solid {
+      children [
+        DEF SHAPE Shape {
+          appearance PBRAppearance {
+            baseColor 0.095395215 0.22841774 0.8000001
+            roughness 1
+            metalness 0
+          }
+          geometry IndexedFaceSet {
```
PeterJCLaw, line 296: Presumably it's not possible to have an `IndexedFaceSet` which declares all 6 exterior faces in one go?

PeterJCLaw (commented):
Nice. I've verified along the major faces of the two inner walls and this seems to give reasonable results now :+1:

PeterJCLaw (approved):
Wrong button, let's ship this :+1:

PeterJCLaw:
Aside: this feels like it could be an excellent candidate for an full-stack test once #271 is merged.

PR #283, opened 2021-01-14
Change zone detection to use controllerArgs and DEF statements

WillB97:
Replaces the previous solution of using robots scene tree index.

This provides a more stable method of detection the zone of a robot, since robots are retrieved by their DEF statement changes to the arena and robot models will not change the reference to the robot.


controllers/territory_controller/territory_controller.py
```diff
@@ -150,9 +150,11 @@ def claim_territory(
             return
 
         new_colour = ZONE_COLOURS[claimed_by]
-        self._robot.getFromDef(station_code).getField("zoneColour").setSFColor(
-            list(new_colour),
-        )
+        station = self._robot.getFromDef(station_code)
```
RealOrangeOne, line 153: When's this `None`?

WillB97: In theory never but since the type is `Optional[Node]` we have to handle it

PeterJCLaw: Presumably it's `None` if the name isn't found?

WillB97: That is true, we should probably throw an error if we can't find a territory.

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -116,16 +116,13 @@ def get_robots(
...
+        robot = supervisor.getFromDef(f"ROBOT-{zone_id}")
         if robot is None:
             if skip_missing:
                 continue
 
-            msg = "Failed to get Webots node for zone {} (id: {})".format(
-                zone_id,
-                webots_id_str,
-            )
+            msg = "Failed to get Webots node for zone {}".format(zone_id)
```
RealOrangeOne, line 125: fffffffffffffff

```suggestion
            msg = f"Failed to get Webots node for zone {zone_id}"
```

RealOrangeOne (approved):

PeterJCLaw (approved):

PeterJCLaw:
Let's hold off merging this until after the leagues though.

controllers/territory_controller/territory_controller.py
```diff
@@ -150,7 +150,12 @@ def claim_territory(
             return
 
         new_colour = ZONE_COLOURS[claimed_by]
-        self._robot.getFromDef(station_code).getField("zoneColour").setSFColor(
+        station = self._robot.getFromDef(station_code)
+        if station is None:
+            raise RuntimeError(
+                f"Failed to fetch territory node {station_code}",
+            )
```
PeterJCLaw, line 157: Maybe make this just a `logging.error`? If this controller crashes out that won't kill the simulation and the error would just as easily be lost in the logs.

Not sure if we think this should be enough to reject a competition match. If it should, then we should maybe consider putting something in place to force that.

PR #286, opened 2021-01-15
League automation script

WillB97:
This script consumes a `league.yaml` file from the compstate repo and runs all the matches listed in the file for a particular arena.
Currently the script only runs the matches and leaves zipping and distributing the logs and ingesting the match files into the compstate.


RealOrangeOne (commented):
There are far more changes in this than I was expecting.

Is it not possible to simply parse the `league.yaml` file, and use that to run the existing `run-comp-match`?

WillB97:
A large proportion of the diff is due to the first commit which just moves the functions out of `run-comp-match` and into the library `match_utils`.
Then subsequent scripts can then just import the library to run matches.

PeterJCLaw:
General comment: the compstate is not a stable format and is not design to be read by arbitrary tools. Please don't consume compstate data directly. Please use the Python `srcomp` library instead. Note that doing so would make it trivial for this to be agnostic of whether it was running league matches or otherwise, as the `srcomp` library provides a uniform interface over all matches.

libraries.txt
```diff
@@ -13,3 +13,4 @@ numpy==1.19.3
 pandas==1.1.4
 scikit-learn==0.23.2
 scipy==1.5.4
+pyyaml==5.3.1
```
PeterJCLaw, line 16: Not sure we should be adding to the competitor libraries because our internal scripts need something.

script/run-comp-match
```diff
@@ -12,5 +12,7 @@ sys.path.insert(1, str(REPO_ROOT / 'modules'))
 import match_utils  # isort:skip
 
 if __name__ == '__main__':
-    match = match_utils.Match()
-    match.main(match.parse_args())
+    try:
+        match_utils.Match()
+    except RuntimeError:
```
PeterJCLaw, line 17: I would discourage catching `RuntimeError`s. All sorts of things can raise this kind of error and generally if you don't know what's causing an error it's better to let it propagate to the caller.

I would also discourage catching errors at this place in the file, especially in this manner. Any error handling which is at the outer script level should probably instead either `exit` itself explicitly. Any other errors should either be converted to useful error messages and/or allowed to bubble so the user gets a traceback they can use to debug. This approach unfortunately masks everything about the error.

script/run-league
```diff
@@ -0,0 +1,96 @@
...
+                str(match_num),
+            ]
+            tlas = [
+                tla if tla is not None else '-'
+                for tla in match[args.arena]
+            ]
+            match_args += tlas
+
+            print(f"Running match {match_num} between {', '.join( tlas)}")
+            match_utils.Match(match_args)
```
PeterJCLaw, line 81: Since these are the same arguments as would be passed to `run-comp-match`, I'm not sure what the point of changing that script was? Surely we could just subprocess it as it was? That would keep this PR much smaller and ensure that each match run got a clean process (as well as keeping it simple to run a single match if needed).

script/run-league
```diff
@@ -0,0 +1,96 @@
...
+        type=Path,
+    )
+    parser.add_argument(
+        'league_file',
+        help=(
+            "The file containing a list of matches to run "
+            "as used by the compstate."
+        ),
+        type=argparse.FileType('r'),
+    )
```
PeterJCLaw, line 37: Reading just the league file is unfortunately not sufficient to correctly know the competitors who should be in a match. I would _strongly_ encourage that either the `srcomp` library or the related CLI package be used to get information about a compstate. Doing otherwise is likely to lead to bugs due to mismatches between how it reads the compstate (which it considers to be internal) and how your implementation does.

As an example, one way this is insufficient is that it doesn't account for teams who have dropped out since the start of the competition. They will still be present in the league file, but not actually expected in the match.

Using `srcomp` itself will be needed for the knockouts anyway as the matches aren't recorded in a file. We may as well therefore do that now and avoid the chances of complex and subtle issues.

modules/match_utils/__init__.py
```diff
@@ -4,14 +4,27 @@
...
+        match_data = self.construct_match_data(args)
+
+        with self.temporary_arena_root(f'match-{match_data.match_number}'):
+            self.prepare_match(args.archives_dir, match_data)
+
+            self.run_match()
+
+            self.collate_logs(args.archives_dir, match_data)
+            self.archive_match_file(args.archives_dir, match_data)
+            self.archive_match_recordings(args.archives_dir)
```
PeterJCLaw, line 26: This is quite a surprising thing for an `__init__` function to do. I would strongly encourage not doing this.
In general it's not clear to me that having this as a class is useful - there's not really any use of it as an object here.

PeterJCLaw (changes requested):
Thanks for having a go at putting this together. I can tell you've put a lot of effort into this.

Unfortunately I don't think this approach is going to work, in part because fundamentally the league match data cannot be read only from the `league.yaml` file, instead SRComp provides an API to get information from a compstate which needs to be used to avoid issues.
It's also not clear to me that the changes here to the existing runner script are sufficient for this approach (there's a lot which seems to still be trying to be command line args, which doesn't make much sense for an in-python API), nor actually necessary in general (we could just subprocess it).

PR #287, opened 2021-01-16
Remove the camera from robots

Adimote:
The camera isn't currently used, lets remove it.


WillB97 (approved):

RealOrangeOne (approved):
Tested there's no more camera, and also no more camera overlay :tada:

PR #288, opened 2021-01-16
Use Claimant name in territory controller log messages

WillB97:
This avoids territories becoming unclaimed printing `BE CLAIMED BY -1 AT 2s`.
An instead prints `BE CLAIMED BY UNCLAIMED AT 2s`.


PeterJCLaw (approved):
:shipit:

PR #291, opened 2021-01-22
Add script to zip logs in archive_dir

WillB97:
Using `./script/zip-comp-logs archives logs --combined` converts
```
archives
├── ABC
│   ├── log-zone-0-match-22.txt
│   ├── log-zone-0-match-23.txt
│   └── log-zone-1-match-24.txt
├── ABC.zip
├── DEF
│   ├── log-zone-1-match-22.txt
│   └── log-zone-1-match-23.txt
├── DEF.zip
├── matches
│   ├── 00x.yaml
│   ├── 022.yaml
│   ├── 023.yaml
│   └── 024.yaml
└── recordings
    ├── match-22.html
    ├── match-22.json
    ├── match-22.mp4
    ├── match-22.x3d
    ├── match-23.html
    ├── match-23.json
    ├── match-23.mp4
    ├── match-23.x3d
    ├── match-24.html
    ├── match-24.json
    ├── match-24.mp4
    ├── match-24.x3d
    └── textures
        └──...
```
to
```
logs
├── combined.zip
├── team-ABC-2021-01-22.zip
└── team-DEF-2021-01-22.zip
```

There are additional options to remove the date from the filename and add a suffix between the TLA and date sections.

The method for identifying which folders are the teams' logs isn't great and simply zips all folders whose name is 3 uppercase letters and an optional number.

### DIFF
--- a/script/zip-comp-logs
+++ b/script/zip-comp-logs
@@ -9,7 +9,7 @@ import argparse
 import datetime
 from typing import List
 from pathlib import Path
-from zipfile import ZipFile
+from zipfile import ZipFile, ZIP_DEFLATED
 
 
 def zip_folder(folder: Path, output_dir: Path, suffix: str, include_date: bool) -> str:
@@ -83,6 +83,17 @@ def parse_args() -> argparse.Namespace:
         action='store_true',
         dest='combined',
     )
+    parser.add_argument(
+        '--animations',
+        help=(
+            "Include the match animation files. "
+            "'all' creates a separate archive of all matches, "
+            "'separate' does the same as 'all' but excludes "
+            "the archive from the combined archive."
+        ),
+        choices=('all', 'separate'),
+        default=None,
+    )
     return parser.parse_args()
 
 
@@ -113,10 +124,32 @@ def main(args: argparse.Namespace) -> None:
             args.include_date,
         ))
 
+    if args.animations in ('all', 'separate'):
+        with ZipFile(
+            args.output_dir / f'animations{suffix_str}.zip',
+            'w',
+            compression=ZIP_DEFLATED,
+        ) as zipfile:
+            for rec_file in (args.archives_dir / 'recordings').iterdir():
+                if rec_file.suffix == '.mp4':
+                    continue
+                zipfile.write(rec_file.resolve(), rec_file.name)
+
+            # save animation textures
+            for texture in (args.archives_dir / 'recordings/textures').glob('**/*'):
+                zipfile.write(
+                    texture.resolve(),
+                    texture.relative_to(args.archives_dir / 'recordings'),
+                )
+
     if args.combined:
         with ZipFile(args.output_dir / f'combined{suffix_str}.zip', 'w') as zipfile:
             for logs_archive in logs_archives:
                 zipfile.write(args.output_dir / logs_archive, logs_archive)
+
+            if args.animations == 'all':
+                animation_zip = f'animations{suffix_str}.zip'
+                zipfile.write(args.output_dir / animation_zip, animation_zip)
 
 
 if __name__ == '__main__':
--- a/script/zip-comp-logs
+++ b/script/zip-comp-logs
@@ -7,11 +7,13 @@ import re
 import shutil
 import argparse
 import datetime
-import tempfile
+import contextlib
+from typing import Union, Iterator
 from pathlib import Path
+from zipfile import ZipFile
 
 
-def zip_folder(folder: Path, args: argparse.Namespace) -> None:
+def zip_folder(folder: Path, args: argparse.Namespace) -> str:
     tla = folder.name
     zip_name = f'team-{tla}'
 
@@ -26,6 +28,7 @@ def zip_folder(folder: Path, args: argparse.Namespace) -> None:
         root_dir=folder,
         base_dir='.',
     )
+    return f'{zip_name}.zip'
 
 
 def is_team_logs_folder(folder: Path) -> bool:
@@ -41,6 +44,15 @@ def is_team_logs_folder(folder: Path) -> bool:
         return True
 
     return False
+
+
+@contextlib.contextmanager
+def optional_zip(filename: Path, create_archive: bool) -> Iterator[Union[None, ZipFile]]:
+    if create_archive:
+        with ZipFile(filename, 'w') as zipfile:
+            yield zipfile
+    else:
+        yield None
 
 
 def parse_args() -> argparse.Namespace:
@@ -81,8 +93,7 @@ def parse_args() -> argparse.Namespace:
     parser.add_argument(
         '--combined',
         help=(
-            "Create a combined archive which contains all the other log archives. "
-            "This is achieved by creating an archive of the contents of the output directory."
+            "Create a combined archive which contains all the other log archives."
         ),
         action='store_true',
     )
@@ -93,30 +104,20 @@ def main(args: argparse.Namespace) -> None:
     # make sure output directory exists
     args.output_dir.mkdir(exist_ok=True)
 
-    for folder in args.archives_dir.iterdir():
-        if not folder.is_dir():
-            # skip files in the archive directory
-            continue
+    with optional_zip(args.output_dir / 'combined2.zip', args.combined) as zipfile:
+        for folder in args.archives_dir.iterdir():
+            if not folder.is_dir():
+                # skip files in the archive directory
+                continue
 
-        if not is_team_logs_folder(folder):
-            # skip unrelated directories
-            continue
+            if not is_team_logs_folder(folder):
+                # skip unrelated directories
+                continue
 
-        zip_folder(folder, args)
+            logs_archive = zip_folder(folder, args)
 
-    if args.combined:
-        # a temporary directory is used since the zip file
-        # can't be created in the directory being archived
-        with tempfile.TemporaryDirectory() as tmpdir_name:
-            # create a combined archive
-            tmpdir_path = Path(tmpdir_name)
-            shutil.make_archive(
-                str(tmpdir_path / 'combined'),
-                'zip',
-                root_dir=args.output_dir,
-                base_dir='.',
-            )
-            shutil.copy(tmpdir_path / 'combined.zip', args.output_dir)
+            if args.combined and isinstance(zipfile, ZipFile):
+                zipfile.write(args.output_dir / logs_archive, logs_archive)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


script/zip-comp-logs
```diff
@@ -0,0 +1,122 @@
...
+    shutil.make_archive(
+        f'{args.output_dir / zip_name}',
+        'zip',
+        root_dir=folder,
+        base_dir='.',
+    )
+
+
+def is_team_logs_folder(folder: Path) -> bool:
+    # Options for detecting which folders are logs:
```
WillB97, line 32: This comment is just listing possible methods of selecting team logs folders, it should probably be removed before merging.

script/zip-comp-logs
```diff
@@ -0,0 +1,122 @@
...
+    # 1. everthing except recordings, matches
+    # 2. folders w/ matching zip file names
+    # 3. regex of folder name
+    # 4. regex of folder contents names
```
WillB97, line 36: This comment is just listing possible methods of selecting team logs folders, it should probably be removed before merging.

script/zip-comp-logs
```diff
@@ -0,0 +1,122 @@
...
+        ),
+        default="",
+    )
+    parser.add_argument(
+        '--no-date',
+        help=(
+            "Remove the date from the archive name."
+        ),
+        action='store_true',
+    )
```
PeterJCLaw, line 79: I suggest making this (at least internally) be expressed in the positive. I forget the exact syntax, but you can do something like:
```suggestion
    parser.add_argument(
        '--no-date',
        dest='include_date',
        help=(
            "Remove the date from the archive name."
        ),
        action='store_false',
        default=True,
    )
```
Such that usage in code is like `args.include_date` (which therefore avoids double-negatives), but the usage in the CLI is still the way round which is probably more obvious.

script/zip-comp-logs
```diff
@@ -0,0 +1,122 @@
...
+    parser.add_argument(
+        '--combined',
+        help=(
+            "Create a combined archive which contains all the other log archives. "
+            "This is achieved by creating an archive of the contents of the output directory."
+        ),
```
PeterJCLaw, line 85: I'm not sure I follow what this means?

Perhaps it would help to (on the PR) provide an example of what the tree inside such an archive would look like?

script/zip-comp-logs
```diff
@@ -0,0 +1,122 @@
...
+        with tempfile.TemporaryDirectory() as tmpdir_name:
+            # create a combined archive
+            tmpdir_path = Path(tmpdir_name)
+            shutil.make_archive(
+                str(tmpdir_path / 'combined'),
+                'zip',
+                root_dir=args.output_dir,
+                base_dir='.',
+            )
+            shutil.copy(tmpdir_path / 'combined.zip', args.output_dir)
```
PeterJCLaw, line 118: Any reason not to just move the file? If the file is large, moving it could be quite a lot faster (as the actual data likely won't need copying).

PeterJCLaw:
Nice, thanks for putting this together. Have left some comments from a quick look.

script/zip-comp-logs
WillB97: I was think this but had forgotten about the dest argument. Solved in 6fe7b2b9062397dbee75803b3c7406aafdb7c836

script/zip-comp-logs
WillB97: The dangers of late night working, however ef3d27c1459c681484f721e29d58cdd77bf18822 makes this redundant as the combined zip is generated differently.

script/zip-comp-logs
WillB97: This was because the combined zip was generated by zipping the entire output directory, not just the zips created while running the script. ef3d27c1459c681484f721e29d58cdd77bf18822 changes this to just zip the files produced by the script.

WillB97: The contents of the combined zip is then:
```console
$ unzip -l combined2.zip 
Archive:  combined2.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
      913  2021-01-23 09:15   team-ABC-2021-01-23.zip
      682  2021-01-23 09:15   team-DEF-2021-01-23.zip
---------                     -------
     1595                     2 files
```

script/zip-comp-logs
```diff
@@ -0,0 +1,124 @@
...
+        action='store_true',
+    )
+    return parser.parse_args()
+
+
+def main(args: argparse.Namespace) -> None:
+    # make sure output directory exists
+    args.output_dir.mkdir(exist_ok=True)
+
+    with optional_zip(args.output_dir / 'combined.zip', args.combined) as zipfile:
```
RealOrangeOne, line 107: This feels like it's the wrong thing to do as a decorator and is probably easier to just do as a block after all the zips have been created, to collect them up

script/zip-comp-logs
```diff
@@ -0,0 +1,124 @@
...
+def is_team_logs_folder(folder: Path) -> bool:
+    # Options for detecting which folders are logs:
+    # 1. everthing except recordings, matches
+    # 2. folders w/ matching zip file names
+    # 3. regex of folder name
+    # 4. regex of folder contents names
+
+    # Using option 3
+    # folder names are three uppercase letters and an optional number
+    if re.match(r'[A-Z]{3}[0-9]*$', folder.name):
```
RealOrangeOne, line 43: I think this regex wants to be slightly different so we can be more sure of its match

```suggestion
    if re.match(r'^[A-Z]{3}[0-9]?$', folder.name):
```
Also, you can just return the output of this directly rather than needing the `if`

RealOrangeOne (commented):
Couple comments, besides that this looks super handy!

script/zip-comp-logs
```diff
@@ -0,0 +1,124 @@
...
+    parser.add_argument(
+        '--no-date',
+        help=(
+            "Remove the date from the archive name."
+        ),
+        dest='include_date',
+        action='store_false',
+    )
+    parser.add_argument(
+        '--combined',
```
RealOrangeOne, line 94: Maybe this should be `--with-combined`? Else it sort of implies it's _only_ creating the combined archive

script/zip-comp-logs
WillB97: `re.match` matches from the start of the string so the `'^'` is implicit, if we want to make that explicit we should probably use `re.search`

PeterJCLaw: It feels a bit odd to me to have an explicit `$` but not a matching `^`. I think it would be reasonable to use `re.match` here, even with explicit `^`.

script/zip-comp-logs
```diff
@@ -0,0 +1,124 @@
...
+import shutil
+import argparse
+import datetime
+import contextlib
+from typing import Union, Iterator
+from pathlib import Path
+from zipfile import ZipFile
+
+
+def zip_folder(folder: Path, args: argparse.Namespace) -> str:
```
PeterJCLaw, line 16: As a general pattern, I would encourage that `args` only be handled at the top level and that any relevant values are passed explicitly into utils. Note also that `argparse.Namespace` isn't usefully typed, so using it here doesn't offer any safety.

script/zip-comp-logs
```diff
@@ -0,0 +1,124 @@
...
+    # Using option 3
+    # folder names are three uppercase letters and an optional number
+    if re.match(r'[A-Z]{3}[0-9]*$', folder.name):
+        return True
+
+    return False
+
+
+@contextlib.contextmanager
+def optional_zip(filename: Path, create_archive: bool) -> Iterator[Union[None, ZipFile]]:
```
PeterJCLaw, line 50: For learnings: `Optional[X]` is generally preferable over `Union[None, X]`.

PeterJCLaw:
What, if anything, is the plan for sharing the animations? I know we've not done that in the past, though we have discussed it and I think it is desirable.
There are probably two approaches:
- we include them in every team's archive, so that they each have a copy (perhaps of only the matches they were in)
- we bundle them separately and send everyone a copy of this extra archive

The latter is the approach which @sedders123 and I designed for the code-submitter when we were planning that that be the distribution mechanism. The primary reason we went that route was that it scales _much_ better to large numbers of teams and match sessions -- we only need to worry about a single copy of the relevant files. (Note that due to the combinatorial nature of things, the data otherwise fairly quickly gets to several GB).

At the moment this script doesn't do anything with the animations, however it does create a "combined" zip, which I assume is intended for upload to some other tool which would then handle the actual distribution to teams?

I think it's worth considering that we don't know what that other tool looks like yet and that there are other things which we may want to have it handle. It's probably therefore simpler to skip the combined zip stuff for now and just do the logs. We can then (either as a separate script or a generalisation of this one) handle bundling the animations zip and subsequently also the integration with the distribution tool.

script/zip-comp-logs
WillB97: 1bd85098719c84177e947c3988163a23ee2eab53

WillB97:
> There are probably two approaches:
> - we include them in every team's archive, so that they each have a copy (perhaps of only the matches they were in)
> - we bundle them separately and send everyone a copy of this extra archive

While the former approach does seem most convenient to teams, the additional space required by a copy of the textures in every archive would make for a much larger overall archive. Equally given the size of the match files (~1MB per match before compression) it's not unreasonable for teams to download an archive of a whole league's matches.

But given the additional storage requirements, even if we decide to distribute each team their matches I would still recommend having a separate animation archive to give to the distribution system. We can then produce the indivual archives at the point of delivery, i.e. [flask can serve a zip file which exists only in memory](https://www.reddit.com/r/flask/comments/6ja1m6/af_in_flask_is_it_possible_to_make_a_zip_file_on/djcqz50?utm_source=share&utm_medium=web2x&context=3).

Another thing to consider is that when we distribute the animations we will need to add some clear documentation of how to view them since Webots' documentation is rather confusing on the matter.

For the moment I'm going to add an option to create an animation archive of all the matches, there will be another PR for adding individual teams' matches to their archive if only to act as reference code for adding the feature to the distribution systems.

script/zip-comp-logs
```diff
@@ -0,0 +1,156 @@
...
+            'w',
+            compression=ZIP_DEFLATED,
+        ) as zipfile:
+            for rec_file in (args.archives_dir / 'recordings').iterdir():
+                if rec_file.suffix == '.mp4':
+                    continue
+                zipfile.write(rec_file.resolve(), rec_file.name)
+
+            # save animation textures
+            for texture in (args.archives_dir / 'recordings/textures').glob('**/*'):
```
RealOrangeOne, line 139:
```suggestion
            for texture in (args.archives_dir / 'recordings' / 'textures').glob('**/*'):
```

RealOrangeOne (approved):

WillB97:
I'm having a look at refactoring the script before merging but I'd like to get srobo/competition-simulator#297 merged so I only have to refactor once.

RealOrangeOne (approved):

PR #292, opened 2021-01-22
Add flag to inhibit movie creation in comp mode

### LINKED ISSUE #290: Add flag to inhibit movie creation in comp mode
Since we are looking at testing code by running short matches in CI and the videos created are discarded, it is unnecessary to record the movie in the first place.

Not doing a recording greatly increases the speed of the simulation and as such decreases the CI time. On an example CPU-only system a 5-second match is reduced from 2 minutes 18 seconds to 24 seconds.

This would most benefit #271 and PeterJCLaw/code-submitter#10

WillB97:
Since we are looking at testing code by running short matches in CI and the videos created are discarded, it is unnecessary to record the movie in the first place.

Not doing a recording greatly increases the speed of the simulation and as such decreases the CI time.

Fixes #290

### DIFF
--- a/controllers/competition_supervisor/competition_supervisor.py
+++ b/controllers/competition_supervisor/competition_supervisor.py
@@ -1,3 +1,4 @@
+import os
 import sys
 import time
 import contextlib
@@ -33,6 +34,11 @@ def record_video(supervisor: Supervisor, file_path: Path) -> Iterator[None]:
     print("Saving video to {}".format(file_path))
 
     config = controller_utils.get_recording_config()
+    if 'SR_COMP_NO_RECORD' in os.environ:
+        print('Not recording movie')
+        yield
+        return
+
     supervisor.movieStartRecording(
         str(file_path),
         width=config.resolution.width,
--- a/controllers/competition_supervisor/competition_supervisor.py
+++ b/controllers/competition_supervisor/competition_supervisor.py
@@ -1,4 +1,3 @@
-import os
 import sys
 import time
 import contextlib
--- a/controllers/competition_supervisor/competition_supervisor.py
+++ b/controllers/competition_supervisor/competition_supervisor.py
@@ -30,13 +30,14 @@ def record_animation(supervisor: Supervisor, file_path: Path) -> Iterator[None]:
 @contextlib.contextmanager
 def record_video(supervisor: Supervisor, file_path: Path) -> Iterator[None]:
     file_path.parent.mkdir(parents=True, exist_ok=True)
-    print("Saving video to {}".format(file_path))
 
     config = controller_utils.get_recording_config()
     if config.quality == 0:
         print('Not recording movie')
         yield
         return
+    else:
+        print("Saving video to {}".format(file_path))
 
     supervisor.movieStartRecording(
         str(file_path),
--- a/controllers/competition_supervisor/competition_supervisor.py
+++ b/controllers/competition_supervisor/competition_supervisor.py
@@ -34,7 +34,7 @@ def record_video(supervisor: Supervisor, file_path: Path) -> Iterator[None]:
     print("Saving video to {}".format(file_path))
 
     config = controller_utils.get_recording_config()
-    if 'SR_COMP_NO_RECORD' in os.environ:
+    if config.quality == 0:
         print('Not recording movie')
         yield
         return


controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -33,6 +34,11 @@ def record_video(supervisor: Supervisor, file_path: Path) -> Iterator[None]:
     print("Saving video to {}".format(file_path))
 
     config = controller_utils.get_recording_config()
+    if 'SR_COMP_NO_RECORD' in os.environ:
```
WillB97, line 37: The alternative of using quality==0, as Peter suggested:
```suggestion
    if config.quality == 0:
```

RealOrangeOne: Personally i'd go for reading the quality, or deriving it from a separate key. Adding an argument to `run_comp_match` a la `--no-record` would then be simple to implement and thus easy to set on CI.

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -33,6 +33,11 @@ def record_video(supervisor: Supervisor, file_path: Path) -> Iterator[None]:
     print("Saving video to {}".format(file_path))
 
     config = controller_utils.get_recording_config()
+    if config.quality == 0:
+        print('Not recording movie')
+        yield
```
RealOrangeOne, line 38: Is the `yield` needed here? Can we just `return`?

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -33,6 +33,11 @@ def record_video(supervisor: Supervisor, file_path: Path) -> Iterator[None]:
     print("Saving video to {}".format(file_path))
```
RealOrangeOne, line 33: It's probably worth wrapping this print in the condition too

PeterJCLaw: Yeah, I'd be tempted to move the config check above the `mkdir` here too. At that point it miiight be worth extracting a separate function which does the context management? Only if that's actually clearer though.

controllers/competition_supervisor/competition_supervisor.py
PeterJCLaw: > Adding an argument to `run_comp_match` a la `--no-record` would then be simple to implement and thus easy to set on CI.

I'm not sure we'd want to set this in CI. Part of the goal of CI was to fully replicate the running of matches to ensure that they weren't broken. Manual inspection of the resulting videos is very much part of that.

If we want to expand the CI running to include other forms of testing (which I hope we do do), then those may not want to have recordings for speed, though they also likely won't be running as competition matches.

I suggest that we not think about this as a CI-first solution -- I don't see running with video recordings as an issue in CI (rather I think it's desirable).

controllers/competition_supervisor/competition_supervisor.py
WillB97: The `yield` is required otherwise this context manager exits immediately and the match is never run

controllers/competition_supervisor/competition_supervisor.py
WillB97: 7689e435504abd036a250cbe9ac547ca5dc3eb73

controllers/competition_supervisor/competition_supervisor.py
WillB97: 5b4cdce1db32995bc1be52ba57e430c8020f25de

controllers/competition_supervisor/competition_supervisor.py
WillB97: > I'm not sure we'd want to set this in CI. Part of the goal of CI was to fully replicate the running of matches to ensure that they weren't broken. Manual inspection of the resulting videos is very much part of that.

I disagree with this, currently the length limitation imposed by the running speed prevents testing actions that don't occur at the start of the match, like the wall controller's functionality. Since the animation is still created, manual inspection of the animation provides the same information as can be gained from the video.

> I suggest that we not think about this as a CI-first solution -- I don't see running with video recordings as an issue in CI (rather I think it's desirable).

But that's what it is, the purpose of this is solely for CI and automated testing.

PeterJCLaw: For doing things for the purpose of testing the environment I think we're going to want a rather different setup. I'm expecting that we'll have a supervisor which exists entirely for testing and that supervisor would be in control of the environment. This would let us do things like test "has the robot moved to the right place after it drove there", as it would be able to inspect the properties of all the nodes. In that setup, having the competition supervisor in place too is going to get very complicated.

How had you imagined the test setup working if you still want the competition supervisor to be controlling things?

WillB97: You've misunderstood me, when testing functionality other than competition mode we wouldn't be running in comp mode so the the discussion of inhibiting video creation is moot. I was talking specifically about our current testing of a comp match in CI, and the possibility of automatically testing of team's code uploads.

My point was that, since we have the animations and we aren't going to be livestreaming these matches, we can still obtain all the information required for testing without needing the video creation.

RealOrangeOne (approved):

PR #293, opened 2021-01-23
Add extra wheel bounding box to achieve full wheel width

WillB97:
This should reduce the occurrence of wheels becoming trapped in towers. Additionally moving to webots-2021a avoids the issue where cylinders are "sucked-in" to other cylinders during collisions.

### DIFF
--- a/protos/Robot_Echo/SRRobot.proto
+++ b/protos/Robot_Echo/SRRobot.proto
@@ -484,7 +484,7 @@ PROTO SRRobot [
                   rotation 0 0 1 1.5708
                   children [
                     Cylinder {
-                      height 0.05
+                      height 0.01
                       radius 0.1
                       subdivision 32
                     }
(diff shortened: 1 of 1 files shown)


protos/Robot_Echo/SRRobot.proto
```diff
@@ -484,7 +484,7 @@ PROTO SRRobot [
                   rotation 0 0 1 1.5708
                   children [
                     Cylinder {
-                      height 0.01
+                      height 0.05
                       radius 0.1
```
RealOrangeOne, line 488: Nit: It'd be great to have these be as close to the size of the wheel as possible, so it looks like they rotate and move correctly.

```suggestion
                      height 0.045
                      radius 0.095
```

image

WillB97: The oversize diameter is so that the wheel doesn't visually sink through the floor once it has settled.

RealOrangeOne (approved):

WillB97:
We are no longer changing the wheel bounding box. Instead we are adding a new, near-weightless, bounding box which increases the width of the wheel.

RealOrangeOne (approved):

PR #295, opened 2021-01-23
Add keyboard control robot code for testing and demos

PeterJCLaw:
This aims to provide a way to explore robot behaviour without needing to write code, for various purposes.

The changes here incorporate the work done in the private repo (specifically https://github.com/srobo/competition-simulator-private/pull/6 and https://github.com/srobo/competition-simulator-private/pull/8), rebased onto the HEAD of this repo to avoid conflicts and remove redundant/unnecessary changes.

Testing for this was manual, it's not obvious how we could test this in an automatic fashion.


controllers/example_controller/keyboard_controller.py
```diff
@@ -0,0 +1,72 @@
+from sr.robot import Robot
+from controller import Keyboard
+
+TIMESTEP = 16
+NO_KEY_PRESSED = -1
+
+
+def print_distance_sensors(robot: Robot) -> None:
+    distance_sensor_names = [
```
RealOrangeOne, line 9: Should we just import this instead?

controllers/example_controller/keyboard_controller.py
```diff
@@ -0,0 +1,72 @@
+from sr.robot import Robot
+from controller import Keyboard
+
+TIMESTEP = 16
```
RealOrangeOne, line 4: We must be able to derive this from somewhere?

controllers/example_controller/keyboard_controller.py
```diff
@@ -0,0 +1,72 @@
...
+
+            elif key == ord('A'):
+                R.motors[0].m0.power = -25
+                R.motors[0].m1.power = 25
+
+            elif key == ord('D'):
+                R.motors[0].m0.power = 25
+                R.motors[0].m1.power = -25
+
+            elif key == ord('E'):
```
RealOrangeOne, line 62: "E" feels annoying to press here - should it just be space?

controllers/example_controller/keyboard_controller.py
PeterJCLaw: I considered that, however they're not the same names.

controllers/example_controller/keyboard_controller.py
PeterJCLaw: Maybe? Bear in mind that this seems to be expected to be different to the basic timestemp. Perhaps @JennyLynnFletcher can provide more context?

WillB97: This is simply how often the control input is sampled.

PeterJCLaw: d0e9162

RealOrangeOne (approved):
This PR as a basis I think looks fine. It'll definitely be useful to use. I'd like to add an API into the `Robot` class itself to support some control like this, but I that's something for a further PR.

PR #297, opened 2021-01-25
Add option for per-team animations

WillB97:
This PR extends the script added in #291
This adds an additional option choice to include teams' match animations in their archives. Since every archive will now include the texture files the resulting combined archive is substantially larger.

The matches to include is determined by extracting the match numbers from the log names, as such matches are added as each log is processed.

`ZIP_DEFLATED` is used for the archives since the animation data is highly compressible and this keeps the overall archive size down.

While it's possible that we choose not to insert the animations into the archives at this point in the chain, this code stands as an example of how team-specific animations can be generated later in the chain.

### DIFF
--- a/script/zip-comp-logs
+++ b/script/zip-comp-logs
@@ -13,7 +13,7 @@ from zipfile import ZipFile, ZIP_DEFLATED
 
 def match_animation_files(log_name: str, archives_dir: Path) -> List[Path]:
     match_num_search = re.search(r'match-([0-9]+)', log_name)
-    if not isinstance(match_num_search, re.Match):
+    if match_num_search is None:
         print(f'Invalid match name: {log_name}')
         return []
     match_num = match_num_search[1]
@@ -45,7 +45,7 @@ def zip_folder(
 
         if animations == 'team':
             # save animation textures
-            for texture in (archives_dir / 'recordings/textures').glob('**/*'):
+            for texture in (archives_dir / 'recordings' / 'textures').glob('**/*'):
                 zipfile.write(
                     texture.resolve(),
                     texture.relative_to(archives_dir / 'recordings/'),
--- a/script/zip-comp-logs
+++ b/script/zip-comp-logs
@@ -4,7 +4,6 @@ A script to zip the logs produced by a series of competition matches.
 """
 
 import re
-import shutil
 import argparse
 import datetime
 from typing import List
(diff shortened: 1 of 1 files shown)


WillB97:
It's also possibly worth keeping this to allow us to generate all types of team zips for manual distribution.

script/zip-comp-logs
```diff
@@ -4,27 +4,53 @@ A script to zip the logs produced by a series of competition matches.
...
 from typing import List
 from pathlib import Path
 from zipfile import ZipFile, ZIP_DEFLATED
 
 
-def zip_folder(folder: Path, output_dir: Path, suffix: str, include_date: bool) -> str:
+def match_animation_files(log_name: str, archives_dir: Path) -> List[Path]:
+    match_num_search = re.search(r'match-([0-9]+)', log_name)
+    if not isinstance(match_num_search, re.Match):
+        print(f'Invalid match name: {log_name}')
```
PeterJCLaw, line 17: Suspect this is going to need a `noqa:T001` unless we've specifically turned it off for this file.

script/zip-comp-logs
```diff
@@ -4,27 +4,53 @@ A script to zip the logs produced by a series of competition matches.
...
 import datetime
 from typing import List
 from pathlib import Path
 from zipfile import ZipFile, ZIP_DEFLATED
 
 
-def zip_folder(folder: Path, output_dir: Path, suffix: str, include_date: bool) -> str:
+def match_animation_files(log_name: str, archives_dir: Path) -> List[Path]:
+    match_num_search = re.search(r'match-([0-9]+)', log_name)
+    if not isinstance(match_num_search, re.Match):
```
PeterJCLaw, line 16: This isn't very pythonic:
```suggestion
    if match_num_search is None:
```

script/zip-comp-logs
```diff
@@ -4,27 +4,53 @@ A script to zip the logs produced by a series of competition matches.
...
+        for log in folder.resolve().iterdir():
+            zipfile.write(log, log.name)
+
+            if animations == 'team':
+                for animation_file in match_animation_files(log.name, archives_dir):
+                    zipfile.write(animation_file.resolve(), animation_file.name)
+
+        if animations == 'team':
+            # save animation textures
+            for texture in (archives_dir / 'recordings/textures').glob('**/*'):
```
PeterJCLaw, line 48:
```suggestion
            for texture in (archives_dir / 'recordings' / 'textures').glob('**/*'):
```

script/zip-comp-logs
```diff
@@ -4,27 +4,53 @@ A script to zip the logs produced by a series of competition matches.
...
+                zipfile.write(
+                    texture.resolve(),
+                    texture.relative_to(archives_dir / 'recordings/'),
```
PeterJCLaw, line 51: Does this need the trailing slash?

WillB97: No, I just tend to leave the trailing slash on directories

PR #303, opened 2021-02-03
Initial CI migration to Github Actions

WillB97:
An attempt to directly convert the CircleCI config to github actions

Current issues
- The build artifacts are additonally zipped
- webots docker image takes several minutes to initialise, on every run

We could partially overcome the first issue by making use of actions/upload-release-asset to directly place the artifact in the release.
~The second issue may be reduced by making a docker image on the github container repository.~


WillB97:
Moving to running Webots directly instead of in docker seems to greatly improve the spin up time to around 80s, all caching attempts I've tried so far have had little benefit on cache-hit and major detriments on misses.

With the build artifacts, moving to automatically deplying them to releases seems like a good solution. With people being able to remove the outer archive if they want the build of a non-tag. A possible solution of just uploading the unarchived build turns out to have shortcomings in that execution permissions are dropped and each file is uploaded as a separate HTTP requests leading to notable overhead.

PeterJCLaw:
Do you have an example of what you mean by the build uploads?
We have one file to upload -- the (custom built) archive of the repo (which I wouldn't expect to preserve execution permissions anyway), so I'm not sure I understand the issues you're hitting?

WillB97:
With github actions artifact even uploading a single zip gets encapsulated within another zip. While you could extract the repo archive and rely on the artifact handling to zip it, you loose the execute permission on all the scripts (which probably isn't a big issue) and the artifact's zipping is significantly slower.

Since the system for uploading files to a release uses a different backend, it doesn't add the encapsulating zip. But I haven't figured out all the syntax for it yet.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,153 @@
...
+        venv/bin/pip install -r all-requirements.txt
+    - name: Lint with flake8
+      run: |
+        source venv/bin/activate
+
+        # Ignore TODOs on CI, for now, even though we do want them
+        # highlighted in development.
+        ./script/linting/lint --extend-ignore=T000
+
+  typecheck:
```
RealOrangeOne, line 94: Given the steps between linting and typechecking are nearly identical, probably makes sense to just combine the steps?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,153 @@
+name: Simulator tests
+
+on:
+  push: {}
+  release:
+    types: [ created ]
```
RealOrangeOne, line 6: Why are we running this when creating releases too?

If the intention is to make publishing better, I'd suggest creating a dedicated workflow for it which only runs on release "publish", and uses https://github.com/RealOrangeOne/date-group/blob/master/.github/workflows/release.yml#L44

This could then be done instead of the archiving stuff below which creates double zips, or I guess as well as if we want to be sure it's working correctly.

.github/workflows/run-match.yml
```diff
@@ -0,0 +1,75 @@
...
+          export DEBIAN_FRONTEND=noninteractive
+          sudo apt-get update
+          sudo apt-get install --yes wget zip pulseaudio
+      - name: Download Webots
+        run: |
+          wget -O ./webots.deb \
+          https://github.com/cyberbotics/webots/releases/download/R${{ matrix.webots-version }}/webots_${{ matrix.webots-version }}_amd64.deb
+      - name: Install Webots & Xvfb
+        run: |
+          sudo apt-get install --yes xvfb ./webots.deb
```
RealOrangeOne, line 35: Probably worth just installing `xvfb` with the installs on line 28? Keep things separate?

.github/workflows/tests.yml
WillB97: We could also merge test and essentially run the check script against all python versions

.github/workflows/run-match.yml
WillB97: Certainly, this was just a hold over from some caching attempts

.github/workflows/tests.yml
WillB97: Is it worth also disabling run-match on releases?

WillB97:
We should probably make a draft release on this branch to test that CI works before merging

.github/workflows/publish.yml
```diff
@@ -0,0 +1,51 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Build archive
+      id: build_archive
+      run: |
+        mkdir archive
+        NAME="competition-simulator-$(git describe --always --tags).zip"
```
RealOrangeOne, line 20: You can probably get the tag from `github.ref` much easier, and then perhaps use it instead of the `set-output` stuff below

.github/workflows/publish.yml
```diff
@@ -0,0 +1,51 @@
+name: Publish simulator archive
+
+on:
+  release:
+    types: [ created ]
```
RealOrangeOne, line 5: You'd be better off using published here, as that means that drafting releases doesn't trigger this, and then it'll be triggered on the final publish. Creating without drafting implies publishing.

```suggestion
    types: [ published ]
```

src: https://github.com/RealOrangeOne/zoloto/blob/de61a0a2068f51a316ba25a6a435336029b6706d/.github/workflows/pypi.yml#L5

.github/workflows/publish.yml
```diff
@@ -0,0 +1,51 @@
...
+    - name: Upload binaries to release
+      uses: svenstaro/upload-release-action@v1-release
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+        file: archive/${{ steps.build_archive.outputs.archive }}
+        asset_name: ${{ steps.build_archive.outputs.archive }}
+        tag: ${{ github.ref }}
+        overwrite: true
+
+  build:
```
RealOrangeOne, line 32: Is there a reason to build _and_ publish here? Once it's in the issue it's probably fine.

.github/workflows/run-match.yml
```diff
@@ -0,0 +1,75 @@
...
+  release:
+    types: [ created ]
+
+jobs:
+  run-comp-match:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+        working-directory: ${{ github.workspace }}
```
RealOrangeOne, line 14: Is this stuff necessary?

.github/workflows/run-match.yml
```diff
@@ -0,0 +1,75 @@
...
+          export DEBIAN_FRONTEND=noninteractive
+          sudo apt-get update
+          sudo apt-get install --yes wget zip pulseaudio xvfb
+      - name: Download Webots
+        run: |
+          wget -O ./webots.deb \
+          https://github.com/cyberbotics/webots/releases/download/R${{ matrix.webots-version }}/webots_${{ matrix.webots-version }}_amd64.deb
+      - name: Install Webots
+        run: |
+          sudo apt-get install --yes ./webots.deb
```
RealOrangeOne, line 35: Skip the `apt-get` wrapper?

```suggestion
          sudo dpkg -i ./webots.deb
```

.github/workflows/run-match.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Prepare team code archive
+        run: |
+          mkdir archives
+          cd archives
+
+          # Convert the sleep comments to print statement so there's a way to
+          # gauge progress and so that Circle doesn't shut things down due to
```
RealOrangeOne, line 42:
```suggestion
          # gauge progress and so that GitHub doesn't shut things down due to
```

README.md
```diff
@@ -125,8 +126,7 @@ zip ABC.zip robot.py
 ## Doing a release
 
 1. Create a new tag & push
-2. Wait for CI to build a zip archive
-3. Upload the archive to the GitHub release for the tag
+2. Wait for CI to build a zip archive and upload it to the tag
```
RealOrangeOne, line 129: Not quite true. On top of pushing a tag, you also need to create a release in GitHub

.github/workflows/tests.yml
```diff
@@ -0,0 +1,69 @@
+name: Simulator tests
+
+on:
+  push: {}
+
+jobs:
+  run-checks:
+    runs-on: ubuntu-latest
```
RealOrangeOne, line 8: We should run on all OS'

.github/workflows/run-match.yml
WillB97: Only when using docker otherwise `/bin/sh` is used

.github/workflows/run-match.yml
PeterJCLaw: Note that `dpkg` won't install dependencies, `apt-get` will do.

.github/workflows/publish.yml
WillB97: We need the tag basename and `github.ref` gives the full ref name

.github/workflows/run-match.yml
WillB97: This was indeed the case

modules/controller_utils/tests.py
```diff
@@ -30,7 +30,7 @@ def fake_tla() -> str:
 
 @contextlib.contextmanager
 def mock_match_file() -> Iterator[IO[str]]:
-    with tempfile.NamedTemporaryFile(suffix='.json', mode='r+t') as f:
+    with tempfile.NamedTemporaryFile(suffix='.json', mode='r+t', delete=False) as f:
```
PeterJCLaw, line 33: This will leave the files around and clutter up developers machines. The point of using the context manager with a temp file is to prevent that. Why is this needed?

WillB97: Windows cannot handle `NamedTemporaryFile` being closed and reopened without it. The files are fractions of KB and will still be removed on reboot.

modules/sr/robot/coordinates/tests.py
```diff
@@ -368,7 +368,9 @@ def test_polar(self) -> None:
 
         for cartesian, expected in cases:
             with self.subTest(cartesian):
-                self.assertEqual(expected, polar_from_cartesian(cartesian))
+                self.assertAlmostEqual(expected.length, polar_from_cartesian(cartesian).length)
+                self.assertAlmostEqual(expected.rot_x, polar_from_cartesian(cartesian).rot_x)
+                self.assertAlmostEqual(expected.rot_y, polar_from_cartesian(cartesian).rot_y)
```
PeterJCLaw, line 373: If we really need to do this please could we avoid the duplicate calls to `polar_from_cartesian`?
However unfortunately this is a semantic change versus the original assertion and one which is not very future-proof.
Suggest that instead of this we perhaps force the rounding ourselves?
```suggestion
                actual = polar_from_cartesian(cartesian)
                # Cope with floating point differences
                rounded = PolarCoord(*(round(x, 7) for x in actual))

                self.assertEqual(expected, actual)
```

modules/controller_utils/tests.py
PeterJCLaw: bcb1959f8953f37714da61eaab03775ab9ab625d

modules/sr/robot/coordinates/tests.py
WillB97: 87bd35c2aa9c3c0e502865d016b27f1251604d0c

.github/workflows/tests.yml
```diff
@@ -0,0 +1,83 @@
...
+      if: steps.cache-dep.outputs.cache-hit != 'true' && matrix.test-os == 'windows-latest'
+      run: |
+        py -m venv venv
+        source venv/Scripts/activate
+        py -m pip install -U pip
+        pip install -r all-requirements.txt
+    - name: Define scripts to test
+      id: define_scripts
+      run: |
+        SCRIPTS=$(find script -type f | grep -vE "/(linting/|testing/|typing/|check$)"|tr "\n" " ")
```
PeterJCLaw, line 58:
```suggestion
        SCRIPTS=$(find script -type f | grep -vE "/(linting/|testing/|typing/|check$)" | tr "\n" " ")
```

PeterJCLaw (approved):

PR #307, opened 2021-02-05
Add script to identify faulty robots in a submission set

WillB97:
Given an extracted archives directory, 1 second matches are run on each archive and failing matches' output is printed.
A summary of the failing robots is also printed at the end.


script/test-robots
```diff
@@ -0,0 +1,70 @@
...
+            try:
+                print(f"Running match for {tla}")  # noqa:T001
+                subprocess.run(
+                    [
+                        sys.executable,
+                        run_comp_match,
+                        str(archives),
+                        '0',  # no logs will be retained so a fixed match number is used
+                        tla,
+                        '-',
```
PeterJCLaw, line 50: Suggest using the utils to gather the number of teams and thus make this future proof against the number of corners changing.

PeterJCLaw:
Niiiice :tada:

It might be worth documenting the various steps that this relies upon to work. The chain of error handling is fairly long here and it feels like it could easily accidentally break. Additionally/alternatively it might be worth a CI test which checks that this does indeed fail for correctly formed zips around bad robot code.

Just to check -- I'm assuming that this _doesn't_ protect against robot code which just takes a long time to get started? (Since then even the competition supervisor would block indefinitely).

WillB97:
> It might be worth documenting the various steps that this relies upon to work. 

Certainly, I'll sort some documentation for it. I might get a chance to add some tests (although probably not for a few days)

> Just to check -- I'm assuming that this _doesn't_ protect against robot code which just takes a long time to get started? (Since then even the competition supervisor would block indefinitely).

Teams which take a long time but don't stall the clock will be caught by the `wait_start` timeout system. As for teams which stall the clock, I've noticed that `subprocess.run` can take a timeout argument so may put in some handling around that.

PeterJCLaw:
> `subprocess.run` can take a timeout argument

How did I not know about this‽

script/test-robots
```diff
@@ -0,0 +1,70 @@
...
+                    ],
+                    check=True,
+                    stdout=subprocess.PIPE,  # we only want to print output from failing robots
+                    stderr=subprocess.STDOUT,
+                )
+
+                print(f"Completed match for {tla}")  # noqa:T001
+
+            except subprocess.CalledProcessError as match_errors:
+                print(match_errors.output.decode())  # noqa: T001
```
PeterJCLaw, line 62: Knowledge sharing: you can pass an `encoding` argument or the `text` flag to `subprocess.run` which causes the encoding be done for you internally.

script/test-robots
WillB97: 5371562d6d4883ed0e6f5cc7d2c3d92bfe8a0c07

script/test-robots
WillB97: Partially implemented in ef15b65cac0493e32218285ea35a8c4eef903938, although the output suplied in the TimeoutExpired error appears to ignore the encoding.

PeterJCLaw: > the output supplied in the TimeoutExpired error appears to ignore the encoding.

This might be worth reporting upstream -- I think this behaviour is surprising enough it might be a bug.

script/test-robots
```diff
@@ -3,13 +3,47 @@
...
 import sys
+import time
 import shutil
 import argparse
 import tempfile
 import subprocess
 from typing import List
 from pathlib import Path
 
+EXCESS_SUBPROCESS_TIMEOUT = 10
```
PeterJCLaw, line 30: A comment here about the unit (or baking that into the constant name) might be useful.

script/test-robots
```diff
@@ -28,22 +28,26 @@ from controller_utils import NUM_ZONES  # isort:skip
...
-def test_webots_runtime(comp_match_script: str, temp_archives: Path) -> int:
-    """
-    Run a empty webots match to determine the expected duration of a successful match.
-    temp_archives are used to recieve the logs and animations.
-    """
+def run_robot_match(
+    comp_match_script: Path,
+    archives_dir: Path,
+    tla: str,
+    failed_tlas: Optional[List[str]],
```
PeterJCLaw, line 37: Is this effectively an output parameter? I would encourage not doing this and instead having this util return a value indicating whether or not the run succeeded.

Python allow returning tuples, so this is fairly easy.

However I would also encourage not returning the duration - that (as far as I can tell) is only used for the calibration and so is a secondary consideration. Moreover it's also something which the caller can trivially measure for themselves.

script/test-robots
```diff
@@ -109,7 +110,13 @@ def main(args: argparse.Namespace) -> None:
...
         tempdir = Path(tempdir_name)
         # Make a temporary copy of the archives to receive the logs
-        archives = shutil.copytree(args.archives_dir.resolve(), tempdir / 'archives')
+        if args.archives_dir.is_dir():
+            archives = shutil.copytree(args.archives_dir.resolve(), tempdir / 'archives')
+        else:
+            archives = tempdir / 'archives'
+            archives.mkdir()
+            with ZipFile(args.archives_dir) as zipfile:
+                zipfile.extractall(archives)
```
PeterJCLaw, line 119: While this is easy to do, it's the sort of complexity I would encourage not having. It complicates the implementation (now there's two things to test) as well as for users (now there's two things I need to understand).
I think I'd stick with the original mechanism here -- just working on a directory of zips since that's the format they need to be in anyway to run the matches.

script/test-robots
```diff
@@ -0,0 +1,141 @@
...
+        ]
+
+        match_args.append(tla)
+
+        match_args.extend(['-'] * (NUM_ZONES - 1))  # all other zones are empty
+
+        match_args.extend([
+            '--duration', '1',
+            '--no-record',
+        ])
```
PeterJCLaw, line 57: The order of the optional (`--named`) arguments doesn't matter -- they could easily be put first to avoid this break I think?
Alternatively (or additionally) we could spray the zones in or just append them:
```suggestion
        num_empty_zones = NUM_ZONES - 1
        match_args = [
            sys.executable,
            str(comp_match_script),
            str(archives_dir),
            '--duration', '1',
            '--no-record',
            '--',
            '0',  # no logs will be retained so a fixed match number is used
            tla,
        ] + (['-'] * num_empty_zones)
```

script/test-robots
```diff
@@ -0,0 +1,141 @@
...
+    failed_tlas: List[str] = []
+
+    run_comp_match = Path(__file__).parent / 'run-comp-match'
+
+    with tempfile.TemporaryDirectory() as tempdir_name:
+
+        tempdir = Path(tempdir_name)
+        # Make a temporary copy of the archives to receive the logs
+        if args.archives_dir.is_dir():
+            archives = shutil.copytree(args.archives_dir.resolve(), tempdir / 'archives')
```
PeterJCLaw, line 114: Is `.resolve()` here useful?

PeterJCLaw (approved):
Nice one :+1: 

Some suggestions inline for improvements.

script/test-robots
WillB97: Not really.

script/test-robots
WillB97: e59a4d998a62a0dea0e8615e63ccbe3f33f4ba9a

script/test-robots
WillB97: 39f318d69424357ecb9c58fd9317ddd1759caac5

script/test-robots
```diff
@@ -115,10 +109,12 @@ def main(args: argparse.Namespace) -> None:
...
 
-        subprocess_timeout = (
-            EXCESS_SUBPROCESS_TIMEOUT +
-            run_robot_match(run_comp_match, archives, '-', None, None)
-        )
+        start_time = time.time()
+        if not run_robot_match(run_comp_match, archives, '-', None):
+            exit(1)
+
+        match_duration = round(time.time() - start_time)
```
PeterJCLaw, line 114: What's the purpose of rounding here?

script/test-robots
WillB97: Removed in 86bea565f705f65a3f0e1cfe2155837501081991

script/test-robots
WillB97: That's a leftover from when I was considering rounding up but timeout is actually a float so there is no need to round.

script/test-robots
```diff
@@ -0,0 +1,133 @@
...
+            tla = team.stem
+
+            # skip non-robot archives
+            if not is_team_archive(tla):
+                continue
+
+            if not run_robot_match(run_comp_match, archives, tla, subprocess_timeout):
+                failed_tlas.append(tla)
+
+    print(f"Failing TLAs: {', '.join(failed_tlas)}")  # noqa: T001
```
RealOrangeOne, line 129: Removing the `f` string here makes this much more readable
```suggestion
    print("Failing TLAs:", ', '.join(failed_tlas))  # noqa: T001
```

script/test-robots
```diff
@@ -0,0 +1,133 @@
...
+        ),
+        type=Path,
+    )
+    return parser.parse_args()
+
+
+def main(args: argparse.Namespace) -> None:
+    failed_tlas: List[str] = []
+
+    run_comp_match = Path(__file__).parent / 'run-comp-match'
```
RealOrangeOne, line 97:
```suggestion
    run_comp_match = REPO_ROOT / 'script' / 'run-comp-match'
```

script/test-robots
```diff
@@ -0,0 +1,133 @@
...
+            str(comp_match_script),
+            str(archives_dir),
+            '--duration', '1',
+            '--no-record',
+            '--',
+            '0',  # no logs will be retained so a fixed match number is used
+            tla,
+        ] + (['-'] * num_empty_zones)
+
+        subprocess.run(
```
RealOrangeOne, line 54: You might like https://docs.python.org/3/library/subprocess.html#subprocess.check_call, which deals with `check=True` for you, and has a slightly nicer return which may be useful.

RealOrangeOne: > Note: Do not use stdout=PIPE or stderr=PIPE with this function. The child process will block if it generates enough output to a pipe to fill up the OS pipe buffer as the pipes are not being read from. 

Yeah best leave as-is

RealOrangeOne (approved):

PR #311, opened 2021-02-11
Fix mass ratio warnings

WillB97:
The extreme mass ratio between the wheel spacer and robot body causes a physics warning to be raised.
SInce the spacer is not attached to the axle and does not rotate it's mass won't effect the movement of the wheel.
This changes from using an arbitrary density to a fixed mass of 1 gram for each spacer.

### DIFF
--- a/protos/Robot_Echo/SRRobot.proto
+++ b/protos/Robot_Echo/SRRobot.proto
@@ -516,7 +516,8 @@ PROTO SRRobot [
           }
           name "right wheel spacer"
           physics Physics {
-            density 0.1
+            density -1
+            mass 0.001
           }
         }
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
Do we know why this avoids the warnings? Presumably the mas previously was in this ballpark anyway, so the difference ratio is still something like 10<sup>5</sup> from my understanding (the wheels being ~10kg I think someone said?)

WillB97:
Currently the spacers are 0.1g compared to 17.3kg for the robot body. The increase of one order of magnitude is enough to remove the warning and it seems like a better idea to set the mass directly instead of using arbitrary densities.

RealOrangeOne (approved):

PR #312, opened 2021-02-11
Stop territory labels obscuring the territory ownership in animations

WillB97:
The output of displays are not rendered in the produced animations. As such the current labels are rendered as an opaque disk over the territory floor so the owner's colour is not visible.

This reduces the display to the area used by the labels so that the territory floor is still visible in the animations. If we want to have the labels visible in the animation outputs we will have to generate the labels as textures.

2021-02-11-200041_1876x698_scrot

### DIFF
--- a/controllers/territory_controller/territory_controller.py
+++ b/controllers/territory_controller/territory_controller.py
@@ -254,7 +254,7 @@ def configure_territory_display(display: Display, station_code: StationCode) -> 
     display.setAlpha(1)
     display.setColor(0x3270ed)
     display.setFont('Arial Black', 48, True)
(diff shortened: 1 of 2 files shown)


WillB97:
Alternatively we can directly use textures for the labels, as in d2954764bacb4d0b5a907ced35eaacbde943d1f9. This has the added benefit of keeping the labels visible in the animations and only increases the texture size by ~50kB.

RealOrangeOne:
50kb is "basically free", especially with #315. Personally i'd vote for #314.

WillB97:
srobo/competition-simulator#314 is a better approach

PR #314, opened 2021-02-15
Convert territory labels to be textures

WillB97:
This is an alternative to srobo/competition-simulator#312 and has the additional upside of making labels visible in the animation.
While this does add 13 more texture files to the animations the total texture size only increases from 3.46MB to 3.50MB.

In videos:
Arena_16

In animations:
image


worlds/Arena.wbt
```diff
@@ -66,77 +66,90 @@ Robot {
       territoryName "PNTerritory"
       emitterName "PNEmitter"
       receiverName "PNReceiver"
+      textureName "../textures/territory-labels/PN.png"
```
RealOrangeOne, line 69: I presume textures can't be SVGs?

RealOrangeOne:
The screenshot in the description makes it look like the territories are now much darker, but I can't replicate that on my machine. Rendering issue?

PeterJCLaw:
> The screenshot in the description makes it look like the territories are now much darker, but I can't replicate that on my machine. Rendering issue?

I'm guessing it's a screenshot of the animation? For me the animations are usually darker.

PeterJCLaw:
Was there a script or something used to create the textures? It would be great if that could be included so that they're easy to edit and rebuild.

worlds/Arena.wbt
WillB97: The Webots docs only mention it supporting PNG and JPG

WillB97:
The labels were made in GIMP while I was waiting for a parcel. Although I should document the settings used but I was unsure where would be the best place for these.

RealOrangeOne (approved):

sedders123 (approved):

PR #315, opened 2021-02-15
Remove unnecessary textures

WillB97:
Reduces animation textures from 3.46MB to 206kB.

Removes roughness, normal and occulusion maps for wheels and changes the floor and background from a texture to fixed colour.

### DIFF
--- a/protos/Robot_Echo/SRRobot.proto
+++ b/protos/Robot_Echo/SRRobot.proto
@@ -439,22 +439,7 @@ PROTO SRRobot [
                     ]
                   }
                   roughness 0.9
-                  roughnessMap ImageTexture {
-                    url [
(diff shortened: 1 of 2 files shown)


RealOrangeOne (approved):
This appears to make very little difference to the visuals on my machine, at least in the viewport. So if it makes that much reduction in the output animation assets, great!

PeterJCLaw:
Hrm, I recall us looking at this in the past and concluding that the tyre textures were fairly noticeable. If that's no longer the case I wonder what's changed?

WillB97:
This doesn't remove the tyre texture but since we are using the `smooth_rubber_base_color` texture the roughness and normal maps change very little.

RealOrangeOne (approved):

sedders123 (approved):

PR #316, opened 2021-02-15
CI improvements

WillB97:
This fixes several shortcomings with the current CI setup:
- Re-enables recording video while running the test match
- Adds support to `run-comp-match` for setting the video resolution (This notabably reduces CI run time)
- Fixes pulseaudio being unavailable, which delayed Webots starting
- Disables camera and display overlays so that the arena is not obscured in the test videos
- Fixes an intermittent bug where Windows doesn't map `python3` to `python`

### DIFF
--- a/script/run-comp-match
+++ b/script/run-comp-match
@@ -33,7 +33,7 @@ def construct_match_data(args: argparse.Namespace) -> controller_utils.MatchData
     ]
 
     recording_config = controller_utils.RecordingConfig(
-        resolution=controller_utils.Resolution(args.resolution[0], args.resolution[1]),
+        resolution=controller_utils.Resolution(*args.resolution),
         quality=100,
     )
 
--- a/script/run-comp-match
+++ b/script/run-comp-match
@@ -32,12 +32,14 @@ def construct_match_data(args: argparse.Namespace) -> controller_utils.MatchData
         for tla in args.tla
     ]
 
-    recording_config = None  # use the default
+    recording_config = controller_utils.RecordingConfig(
+        resolution=controller_utils.Resolution(args.resolution[0], args.resolution[1]),
+        quality=100,
+    )
+
     if args.no_record:
-        recording_config = controller_utils.RecordingConfig(
-            resolution=controller_utils.Resolution(0, 0),  # the resolution is unused
-            quality=0,  # signal to the competition_supervisor to not generate a video
-        )
+        # signal to the competition_supervisor to not generate a video
+        recording_config = recording_config._replace(quality=0)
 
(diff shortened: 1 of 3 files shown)


script/run-comp-match
```diff
@@ -32,12 +32,14 @@ def construct_match_data(args: argparse.Namespace) -> controller_utils.MatchData
         for tla in args.tla
     ]
 
-    recording_config = None  # use the default
+    recording_config = controller_utils.RecordingConfig(
+        resolution=controller_utils.Resolution(args.resolution[0], args.resolution[1]),
+        quality=100,
```
RealOrangeOne, line 37: Maybe just make this `0 if args.no_record else 100`, and save the mutation below?

script/run-comp-match
```diff
@@ -232,6 +234,14 @@ def parse_args() -> argparse.Namespace:
         action='store_true',
         dest='no_record',
     )
+    parser.add_argument(
+        '--resolution',
+        help="Set the resolution of the produced video. The format is <width> <height>",
```
PeterJCLaw, line 239: Specifying the format is redundant -- `argparse` will include that for you:
```
$ ./script/run-comp-match 
usage: run-comp-match [-h] [--duration DURATION] [--no-record]
                      [--resolution width height]
                      archives_dir match_num tla tla
run-comp-match: error: the following arguments are required: archives_dir, match_num, tla

$ ./script/run-comp-match -h
  ...
  --resolution width height
                        Set the resolution of the produced video. The format
                        is <width> <height>
```

```suggestion
        help="Set the resolution of the produced video.",
```

script/run-comp-match
```diff
@@ -32,12 +32,14 @@ def construct_match_data(args: argparse.Namespace) -> controller_utils.MatchData
         for tla in args.tla
     ]
 
-    recording_config = None  # use the default
+    recording_config = controller_utils.RecordingConfig(
+        resolution=controller_utils.Resolution(args.resolution[0], args.resolution[1]),
```
PeterJCLaw, line 36: I'd be tempted to change this to spray the arguments in:
```suggestion
        resolution=controller_utils.Resolution(*args.resolution),
```

script/run-comp-match
WillB97: Fair point

script/run-comp-match
WillB97: 28f292460e8219448d57b8c5236ea0c02fafa175

sedders123 (approved):

RealOrangeOne (approved):

PR #318, opened 2021-02-24
Lint proto files

WillB97:
Fixes whitespace in the proto files as well as removing excessive decimal places.


PeterJCLaw:
How was this linting done? Can we get the tool run on CI?

WillB97:
Manually, yet to find anything that actually understands that format.

protos/Robot_Echo/SRRobot.proto
```diff
@@ -14,980 +14,979 @@ PROTO SRRobot [
...
-          ]
-          name "front bump sensor"
-          boundingObject Transform {
-            translation 0 0 0
+    children [
+      DEF bump_sensors Group {
+        children [
+          TouchSensor {
+            translation 0 0.09 -0.03
+            rotation 0 1 0 -3.14
```
PeterJCLaw, line 22: not sure this float reduction is desirable? It looks a lot like the value here is π, which we probably don't want to truncate.

protos/Robot_Echo/SRRobot.proto
```diff
@@ -14,980 +14,979 @@ PROTO SRRobot [
...
+              2 1000 0
+            ]
           }
-        }
-        Solid {
-          translation 0.19999998777297837 0.00026863610916181684 -0.00036030151878293246
-          rotation -0.9999999953968224 4.284946735166644e-05 8.585032564057752e-05 0.9344151591877564
-          boundingObject DEF BOUNDING_WHEEL Group {
+          DEF front_right DistanceSensor {
+            translation 0.12 0.17 -0.01
```
PeterJCLaw, line 277: I'm not sure we should be moving the distance sensors or the joints or any physical attributes of the robot which might impact its behaviour.

PeterJCLaw (changes requested):
Thanks for putting this together.

:+1: to the whitespace fixing and to sorting out the LEDs, however
:-1: on the changes to the structure/position of non-aesthetic components of the robot

protos/Robot_Echo/SRRobot.proto
WillB97: The arena is only rotated north to 4 d.p. precision so it seems unnecessary to have this rotation to 16 d.p., ultimately this is a change of under 0.1 degrees.

protos/Robot_Echo/SRRobot.proto
```diff
@@ -339,8 +339,8 @@ PROTO SRRobot [
               }
             ]
             endPoint Solid {
-              translation 0.19999998777297837 0.00026863610916181684 -0.00036030151878293246
-              rotation -0.9999999953968224 4.284946735166644e-05 8.585032564057752e-05 0.9344151591877564
+              translation 0.2 0 0
+              rotation -1 0 0 0
```
WillB97, line 343: The previous rotation here just rotated the wheel approximately around its axis. In fact the axis used leaves the wheels pointing fractionally inward.

WillB97:
We now have a script to both test and fix most of the whitespace linting.

script/proto-lint
```diff
@@ -0,0 +1,116 @@
...
+                    # append this line to the previous line
+                    new_lines[-1] = new_lines[-1][:-1] + new_line + '\n'
+                    last_line_bracket = ''
+                    continue
+
+            if current_indent < 0:
+                logging.error('Brackets mismatch, indentation went negative. Aborting')
+                return False
+
+            if new_line != "":
```
PeterJCLaw, line 51: nit: I think this can be reduced to:
```suggestion
            if new_line:
```

PeterJCLaw (commented):
Code looks good, some suggestions for improvement.

I remain concerned that some of the changes to the proto files could have actual (albeit small) impacts to the robot/arena behaviour.

PeterJCLaw:
> I remain concerned that some of the changes to the proto files could have actual (albeit small) impacts to the robot/arena behaviour.

Perhaps we could break this into two PRs:
- one which introduces the script and its effects (plus maybe the other manual file layout changes?)
- a follow-on PR which adapts the numbers etc.

PR #320, opened 2021-02-27
Territories which are disconnected don't count towards breakage

WillB97:
Fix the bug

### DIFF
--- a/controllers/territory_controller/territory_controller.py
+++ b/controllers/territory_controller/territory_controller.py
@@ -120,9 +120,9 @@ class ClaimLog:
 
     def get_claim_count(self, station_code: StationCode) -> int:
         return len([
-            x
-            for x, y, _, _ in self._log
-            if x == station_code and y != Claimant.UNCLAIMED
+            log_station
+            for log_station, claimant, _, _ in self._log
+            if log_station == station_code and claimant != Claimant.UNCLAIMED
         ])
 
     def _record_log_entry(self, entry: ClaimLogEntry) -> None:
--- a/controllers/territory_controller/territory_controller.py
+++ b/controllers/territory_controller/territory_controller.py
@@ -119,7 +119,11 @@ class ClaimLog:
         return station_code in self._locked_territories
 
     def get_claim_count(self, station_code: StationCode) -> int:
-        return len([x for x, _, _, _ in self._log if x == station_code])
+        return len([
+            x
+            for x, y, _, _ in self._log
+            if x == station_code and y != Claimant.UNCLAIMED
+        ])
 
     def _record_log_entry(self, entry: ClaimLogEntry) -> None:
         self._log.append(entry)
@@ -335,7 +339,10 @@ class TerritoryController:
             )
             return
 
-        if self._claim_log.get_claim_count(station_code) == LOCKED_OUT_AFTER_CLAIM - 1:
+        if (
+            self._claim_log.get_claim_count(station_code) == LOCKED_OUT_AFTER_CLAIM - 1 and
+            claimed_by != Claimant.UNCLAIMED
+        ):
             # This next claim would trigger the "locked out" condition, so rather than
             # making the claim, instead cause a lock-out.
             set_node_colour(station, LOCKED_COLOUR)
--- a/controllers/territory_controller/territory_controller.py
+++ b/controllers/territory_controller/territory_controller.py
@@ -351,7 +351,7 @@ class TerritoryController:
 
         if (
             self._claim_log.get_claim_count(station_code) == LOCKED_OUT_AFTER_CLAIM - 1 and
-            claimed_by != Claimant.UNCLAIMED
+            claimed_by in Claimant.zones()
         ):
             # This next claim would trigger the "locked out" condition, so rather than
             # making the claim, instead cause a lock-out.
--- a/controllers/territory_controller/territory_controller.py
+++ b/controllers/territory_controller/territory_controller.py
@@ -33,12 +33,12 @@ class Claimant(enum.IntEnum):
     ZONE_1 = 1
 
     @classmethod
-    def zones(cls) -> Tuple[Claimant, ...]:
-        return tuple(
+    def zones(cls) -> Set[Claimant]:
+        return {
             claimant
             for claimant in cls
             if claimant != cls.UNCLAIMED
-        )
+        }
 
 
 POINTS_PER_TERRITORY = 2
--- a/controllers/territory_controller/territory_controller.py
+++ b/controllers/territory_controller/territory_controller.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import sys
 import enum
 import struct
@@ -29,6 +31,14 @@ class Claimant(enum.IntEnum):
     UNCLAIMED = -1
     ZONE_0 = 0
     ZONE_1 = 1
+
+    @classmethod
+    def zones(cls) -> Tuple[Claimant, ...]:
+        return tuple(
+            claimant
+            for claimant in cls
+            if claimant != cls.UNCLAIMED
+        )
 
 
 POINTS_PER_TERRITORY = 2
@@ -122,7 +132,7 @@ class ClaimLog:
         return len([
             log_station
             for log_station, claimant, _, _ in self._log
-            if log_station == station_code and claimant != Claimant.UNCLAIMED
+            if log_station == station_code and claimant in Claimant.zones()
         ])
 
     def _record_log_entry(self, entry: ClaimLogEntry) -> None:
--- a/controllers/territory_controller/tests.py
+++ b/controllers/territory_controller/tests.py
@@ -1,7 +1,7 @@
 import re
 import unittest
 from pathlib import Path
-from unittest.mock import Mock, patch
+from unittest.mock import patch
 
 from territory_controller import (
     Claimant,
@@ -205,12 +205,7 @@ class TestTerritoryLockout(unittest.TestCase):
     @patch('controller.Supervisor.getFromDef')
     @patch('territory_controller.get_robot_device')
     @patch('territory_controller.TerritoryController.set_score_display')
-    def setUp(
-        self,
-        mock_score_display: Mock,
-        mock_get_robot: Mock,
-        mock_getFromDef: Mock,
-    ) -> None:
+    def setUp(self, _: object, __: object, ___: object) -> None:
         super().setUp()
         claim_log = ClaimLog(record_arena_actions=False)
         self.load_territory_owners(claim_log)
@@ -221,7 +216,7 @@ class TestTerritoryLockout(unittest.TestCase):
         )
 
     @patch('controller.Supervisor.getFromDef')
-    def test_territory_lockout(self, mock_getFromDef: Mock) -> None:
+    def test_territory_lockout(self, _: object) -> None:
         """
         Test a territory is locked after the correct number of claims and
         disconnected territories aren't also locked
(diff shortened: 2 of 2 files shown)


controllers/territory_controller/territory_controller.py
```diff
@@ -119,7 +119,11 @@ def is_locked(self, station_code: StationCode) -> bool:
         return station_code in self._locked_territories
 
     def get_claim_count(self, station_code: StationCode) -> int:
-        return len([x for x, _, _, _ in self._log if x == station_code])
+        return len([
+            x
+            for x, y, _, _ in self._log
+            if x == station_code and y != Claimant.UNCLAIMED
```
PeterJCLaw, line 125: Feels like the positive test here would be clearer, especially given what the but was:
```suggestion
            if x == station_code and y == Claimant.CLAIMED
```

PeterJCLaw: Except that `Claimant.CLAIMED` isn't a thing...

PeterJCLaw:
Would be great to have some tests for this.

controllers/territory_controller/territory_controller.py
```diff
@@ -119,7 +119,11 @@ def is_locked(self, station_code: StationCode) -> bool:
         return station_code in self._locked_territories
 
     def get_claim_count(self, station_code: StationCode) -> int:
-        return len([x for x, _, _, _ in self._log if x == station_code])
+        return len([
+            x
+            for x, y, _, _ in self._log
```
RealOrangeOne, line 124: Descriptive variable names here would be good!

WillB97: eae94f394851bd0c2ceb536b4fcfea4d4fd94b6a

controllers/territory_controller/territory_controller.py
WillB97: 5f124e202bcf7c6f8cdce7d3668beccc4651859d

controllers/territory_controller/territory_controller.py
```diff
@@ -30,6 +32,14 @@ class Claimant(enum.IntEnum):
     ZONE_0 = 0
     ZONE_1 = 1
 
+    @classmethod
+    def zones(cls) -> Tuple[Claimant, ...]:
```
PeterJCLaw, line 36: I suspect consumers don't need to know this is a tuple -- I'd be tempted to use `Sequence` or even just `Collection` instead.

controllers/territory_controller/territory_controller.py
```diff
@@ -335,7 +349,10 @@ def set_territory_ownership(
             )
             return
 
-        if self._claim_log.get_claim_count(station_code) == LOCKED_OUT_AFTER_CLAIM - 1:
+        if (
+            self._claim_log.get_claim_count(station_code) == LOCKED_OUT_AFTER_CLAIM - 1 and
+            claimed_by != Claimant.UNCLAIMED
```
PeterJCLaw, line 354: Maybe also update this one?
```suggestion
            claimed_by in Claimant.zones()
```

controllers/territory_controller/territory_controller.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
RealOrangeOne, line 1: Is this import needed? Feels a bit suspect

WillB97: Mypy won't let you have a classmethod which returns its own class in the type signature without it. See line 36

PeterJCLaw: The more usual (though perhaps now "legacy") way to solve that would be to quote the annotation.
I'd not realised that this future import was available to us though :)

WillB97: The feature is added in 3.10

RealOrangeOne: Ah yeah I've seen it as quoted, and I knew newer versions supported just specifying it. Yeah let's keep the future import then :+1:

RealOrangeOne (approved):
Thoughts on adding a test case for this?

controllers/territory_controller/territory_controller.py
WillB97: e0eabbecc0093b423489414f1a4bcce1964c9dcf

WillB97:
While having a test case may have prevented this issue in the first place, since we have now fixed this and the lock-out system isn't going to have any changes before it's removal at the end of this year the addition of a test case at this point seems of little benefit.

PeterJCLaw:
> While having a test case may have prevented this issue in the first place, since we have now fixed this and the lock-out system isn't going to have any changes before it's removal at the end of this year the addition of a test case at this point seems of little benefit.

I appreciate this that it feels like this may not change, however I'm not sure this is the right approach:
- experience shows that things change much more than you'd think and often in ways that you don't predict. The number of times I've thought "nah, this won't change, I don't need to write tests" is nonzero. The number of times I've been right about that assumption is zero. 
- as well as helping developers have confidence in correct behaviour, tests help readers understand the expected behaviour of the code and how it should be used

WillB97:
I've added tests for cut-off territories and locking territories through self-claiming

prophile:
New territory names (assignment your discretion):

* FL: Flailsbury
* YL: Yamlot
* HA: Hattersly
* TH: Tudor Hammer
* PL: Port Lucent
* SF: Scarborough Fire

controllers/territory_controller/tests.py
```diff
@@ -184,3 +187,85 @@ def test_BE_links(self) -> None:
...
+                    self.territory_controller._claim_log.is_locked(station),
+                    f"Territory {station.value} locked early on claim {i}",
+                )
+
+            # capture BE, SZ by opposing side
+            claimant = Claimant(i % 2)
+            for station in [StationCode.BE, StationCode.SZ]:
+                # when BE becomes locked SZ will be uncapturable
+                # which will prevent it from also locking
+                self.territory_controller.claim_territory(station, claimant, 0)
```
PeterJCLaw, line 240: Please can we make this explicit about which claimant rather than being clever about it.

controllers/territory_controller/tests.py
```diff
@@ -184,3 +187,85 @@ def test_BE_links(self) -> None:
...
+
+
+class TestTerritoryLockout(unittest.TestCase):
+    "Test that individual territories become 'locked' when claimed a set number of times"
+
+    _zone_0_territories = {StationCode.PN, StationCode.EY}
+    _zone_1_territories = {StationCode.YL, StationCode.PO}
+
+    def load_territory_owners(self, claim_log: ClaimLog) -> None:
+        # set territories owned by zone 0
```
PeterJCLaw, line 199:
```suggestion
```

controllers/territory_controller/tests.py
```diff
@@ -184,3 +187,85 @@ def test_BE_links(self) -> None:
...
+        for territory in self._zone_0_territories:
+            claim_log._station_statuses[territory] = Claimant.ZONE_0
+
+        # set territories owned by zone 1
```
PeterJCLaw, line 203:
```suggestion
```

controllers/territory_controller/tests.py
WillB97: The idea of the loop is to alternate claims between teams since testing a team repeatedly claiming a territory is a different test. See 7eab2ec7cce443601227e50ac9594ca46c21fd7d

controllers/territory_controller/tests.py
```diff
@@ -184,3 +187,85 @@ def test_BE_links(self) -> None:
...
+            claim_log._station_statuses[territory] = Claimant.ZONE_1
+
+    @patch('controller.Supervisor.getFromDef')
+    @patch('territory_controller.get_robot_device')
+    def setUp(self, mock_get_robot: Mock, mock_getFromDef: Mock) -> None:
+        super().setUp()
+        claim_log = ClaimLog(record_arena_actions=False)
+        self.load_territory_owners(claim_log)
+        self.attached_territories = AttachedTerritories(claim_log)
+        with patch('territory_controller.TerritoryController.set_score_display'):
```
PeterJCLaw, line 212: Would be great to be consistent about how we're doing this patching -- either all as context managers or all as decorators.

controllers/territory_controller/tests.py
```diff
@@ -204,16 +204,21 @@ def load_territory_owners(self, claim_log: ClaimLog) -> None:
 
     @patch('controller.Supervisor.getFromDef')
     @patch('territory_controller.get_robot_device')
-    def setUp(self, mock_get_robot: Mock, mock_getFromDef: Mock) -> None:
+    @patch('territory_controller.TerritoryController.set_score_display')
+    def setUp(
+        self,
+        mock_score_display: Mock,
+        mock_get_robot: Mock,
+        mock_getFromDef: Mock,
```
PeterJCLaw, line 212: Since we're not using these I'd probably call them `_`, `__` and `___` (they'll need to be unique I think) and give then all an annotation of `object`. (also applies to the ones on the tests).

controllers/territory_controller/tests.py
```diff
@@ -184,3 +187,91 @@ def test_BE_links(self) -> None:
...
+
+            # capture BE, SZ by alternating sides
+            if i % 2 == 0:
+                claimant = Claimant.ZONE_0
+            else:
+                claimant = Claimant.ZONE_1
+            for station in [StationCode.BE, StationCode.SZ]:
+                # when BE becomes locked SZ will be uncapturable
+                # which will prevent it from also locking
+                self.territory_controller.claim_territory(station, claimant, 0)
```
PeterJCLaw, line 246: Sorry, I should have been clearer about this -- my concern here is that using things like modulo to determine behaviours creates complex implicit couplings between the conditional block and the loop iterations. This is generally a bad idea in live code and highly undesirable in tests. In tests, I strongly discourage the use of conditionals and all but the simplest of loops.

If this does need to be loops then something like this is probably clearer:
```python
        for claimant, station in (
            (Claimant.ZONE_0, StationCode.BE),
            (Claimant.ZONE_1, StationCode.SZ),
        ):
            with self.subTest(claimant=claimant, station=station):
                for i in range(LOCKED_OUT_AFTER_CLAIM):
                    self.assertFalse(
                        self.territory_controller._claim_log.is_locked(station),
                        f"Territory {station.value} locked early on claim {i}",
                    )

                self.territory_controller.claim_territory(station, claimant, 0)
```
Though I'd be tempted to just inline the calls.

controllers/territory_controller/tests.py
WillB97: 21740203e9f877176115e141713ac0e64a38fd5c

controllers/territory_controller/tests.py
WillB97: 682cb6e0d4248ddaa454d0e4227e9ef15b3d8ed5

controllers/territory_controller/tests.py
PeterJCLaw: https://github.com/srobo/competition-simulator/pull/320/commits/0c3292a16be1d0eeb2de0b44a9c353460a796bce

controllers/territory_controller/tests.py
PeterJCLaw: (discussion happened, then https://github.com/srobo/competition-simulator/pull/320/commits/0c3292a16be1d0eeb2de0b44a9c353460a796bce)

PeterJCLaw (approved):

RealOrangeOne (approved):

PR #321, opened 2021-02-27
Multiplayer keyboard controls

RealOrangeOne:
- Add a separate key map so both robots can be driven simultaneously
- Add a modifier to increase speed of forwards/backwards movement

### DIFF
--- a/controllers/example_controller/keyboard_controller.py
+++ b/controllers/example_controller/keyboard_controller.py
@@ -5,12 +5,13 @@ KEYBOARD_SAMPLING_FREQUENCY = 16
 NO_KEY_PRESSED = -1
 
 CONTROLS = {
-    "forward": ("W", "I"),
-    "reverse": ("S", "K"),
-    "left": ("A", "J"),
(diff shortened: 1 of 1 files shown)


controllers/example_controller/keyboard_controller.py
```diff
@@ -43,26 +61,34 @@ def print_distance_sensors(robot: Robot) -> None:
...
+                R.motors[0].m0.power = 100
+                R.motors[0].m1.power = 100
+
+            elif key == key_reverse:
+                R.motors[0].m0.power = -50
+                R.motors[0].m1.power = -50
+
+            elif key == key_boost + key_reverse:
                 R.motors[0].m0.power = -50
                 R.motors[0].m1.power = -50
```
sedders123, line 78: Should these not be -100 as we're in boost mode?
```suggestion
                R.motors[0].m0.power = -100
                R.motors[0].m1.power = -100
```

sedders123 (approved):

PR #323, opened 2021-03-03
Attempt to clarify this test

PeterJCLaw:


controllers/territory_controller/tests.py
```diff
@@ -216,40 +231,34 @@ def setUp(self, _: object, __: object, ___: object) -> None:
...
+        self.claim_territory(StationCode.PN, Claimant.ZONE_1)
+        self.claim_territory(StationCode.EY, Claimant.ZONE_1)
+
+        self.assertNotLocked(StationCode.PN, "early after second claims")
+        self.assertNotLocked(StationCode.EY, "early after second claims")
+
+        self.claim_territory(StationCode.PN, Claimant.ZONE_0)
+
+        self.assertLocked(StationCode.PN, "after third claim")
+        self.assertNotLocked(StationCode.EY, "after third claim")
```
WillB97, line 261: We haven't actually claimed this
```suggestion
        self.assertNotLocked(StationCode.EY, "after lock-out of PN")
```

controllers/territory_controller/tests.py
```diff
@@ -216,40 +231,34 @@ def setUp(self, _: object, __: object, ___: object) -> None:
...
+    @patch('territory_controller.TERRITORY_LINKS', new={
+        (StationCode.PN, StationCode.EY),
+        (TerritoryRoot.z0, StationCode.PN),
+        (TerritoryRoot.z1, StationCode.PN),
+    })
     def test_territory_lockout(self, _: object) -> None:
         """
         Test a territory is locked after the correct number of claims and
         disconnected territories aren't also locked
         """
```
WillB97, line 244:
```suggestion
        EY is only capturable when PN is owned and
        has its ownership reset when PN's owner changes
        """
```

WillB97 (approved):
While this hardcodes the `LOCKED_OUT_AFTER_CLAIM` value it makes things more obvious

controllers/territory_controller/tests.py
PeterJCLaw: Does 8d30602 convey this?

WillB97: Yes, that works.

WillB97 (approved):

PeterJCLaw:
> While this hardcodes the `LOCKED_OUT_AFTER_CLAIM` value it makes things more obvious

How do you mean? The test also mocks out the value, so the coupling is all within the test itself.

PR #324, opened 2021-03-07
Use runpy rather than exec

PeterJCLaw:
The [`runpy` package](https://docs.python.org/3/library/runpy.html) is what implements `python -m X.py` and `python X.py` running, so it's pretty much designed for what we're trying to do here.

The main reason to switch to it is that it preserves the original file name of the code being run and thus plays much better with debuggers.

### DIFF
--- a/controllers/sr_controller/sr_controller.py
+++ b/controllers/sr_controller/sr_controller.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import runpy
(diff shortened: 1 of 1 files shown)


WillB97 (approved):
This seems to work. Stacktraces also print the correct name now which should make tracking down bugs across multiple source files much better.

PR #325, opened 2021-03-08
Improve ClaimLog class data format

WillB97:
Converts `ClaimLogEntry` to `NamedTuple` and combines `_station_statuses` and `_locked_territories` into a single `dataclass`.

### DIFF
--- a/controllers/territory_controller/territory_controller.py
+++ b/controllers/territory_controller/territory_controller.py
@@ -5,7 +5,7 @@ import enum
 import struct
 import logging
 import collections
-from typing import Set, Dict, List, Tuple, Union, Mapping, NamedTuple
+from typing import Set, Dict, List, Tuple, Union, Mapping
 from pathlib import Path
 from collections import defaultdict
 from dataclasses import dataclass
@@ -113,7 +113,8 @@ class StationStatus:
     locked: bool = False
 
 
-class ClaimLogEntry(NamedTuple):
+@dataclass(frozen=True)
+class ClaimLogEntry:
     station_code: StationCode
     claimant: Claimant
     claim_time: float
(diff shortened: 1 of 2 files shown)


controllers/territory_controller/territory_controller.py
```diff
@@ -106,33 +107,45 @@ class TerritoryRoot(str, enum.Enum):
 }
 
 
-ClaimLogEntry = Tuple[StationCode, Claimant, float, bool]
+@dataclass
+class StationStatus:
+    owner: Claimant = Claimant.UNCLAIMED
+    locked: bool = False
+
```
RealOrangeOne, line 114: Using both a `dataclass` and `NamedTuple` here concerns me, but I'm not sure why. Is there a reason it's a `dataclass`?

controllers/territory_controller/territory_controller.py
```diff
@@ -106,33 +107,45 @@ class TerritoryRoot(str, enum.Enum):
...
+
+class ClaimLogEntry(NamedTuple):
```
PeterJCLaw, line 116: What impact does this have on the output format, if any?

PeterJCLaw: nevermind, I forgot we explicitly construct the output.

PeterJCLaw (approved):

controllers/territory_controller/territory_controller.py
PeterJCLaw: Presumably because it needs to be mutable?

If anything I'd perhaps suggest making the other also be a `dataclass`, though maybe adding `frozen=True` if we want it immutable?

WillB97: 51e27b73624e907c64e5943f74a1e10dbb089397

sedders123 (approved):

RealOrangeOne (approved):

PR #326, opened 2021-03-09
CI tweaks

WillB97:
This limits CI to running on PRs and the `master` branch and auto cancels redundant builds.
i.e. if you push another commit before the previous workflow has completed the old workflow is cancelled. Also when a new commit is created which adds no new content over previous commits (merge commit, etc.) the CI is skipped with a link to the completed run.

Currently running on PRs occurs when:
- the pull request is created
- commit(s) are pushed to the pull request
- a closed pull request is reopened

### DIFF
--- a/.github/workflows/run-match.yml
+++ b/.github/workflows/run-match.yml
@@ -1,7 +1,10 @@
 name: Simulator match
 
 on:
-  push: {}
+  push:
+    branches:
+      - master
+  pull_request: {}
   release:
     types: [ created ]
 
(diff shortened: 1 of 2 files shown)


.github/workflows/run-match.yml
```diff
@@ -1,12 +1,26 @@
...
 
 jobs:
+  skip_duplicates:
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@v3.3.0
```
RealOrangeOne, line 19: Any reason to use this over https://github.com/styfle/cancel-workflow-action? It seems to be used more

WillB97: As far I can tell that one does not handle skipping commits with the same information, only cancelling previous runs.

PR #327, opened 2021-03-09
Change publish workflow to auto-create releases from tags

WillB97:
The release name uses the tag name and the tag annotation is used for the release text.

This also still supports creating releases from the Github UI as long as it doesn't use an existing tag.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -9,9 +9,8 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-      with:
-        fetch-depth: 0
+    # using v2 strips tag annotations https://github.com/actions/checkout/issues/290
+    - uses: actions/checkout@v1
     - name: Set up Python 3.7
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,14 +1,16 @@
 name: Publish simulator archive
 
 on:
-  release:
-    types: [ published ]
+  push:
+    tags:
+      - '**'
 
 jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    # checkout@v1 must be used since v2 breaks tags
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -1,14 +1,16 @@
 name: Publish simulator archive
 
 on:
-  release:
-    types: [ published ]
+  push:
+    tags:
+      - '**'
```
RealOrangeOne, line 6: Is this meant to be double?

```suggestion
      - '*'
```

.github/workflows/publish.yml
```diff
@@ -1,14 +1,16 @@
...
 
 jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    # checkout@v1 must be used since v2 breaks tags
+    - uses: actions/checkout@v1
```
RealOrangeOne, line 13: > Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags. Refer here to learn which commit $GITHUB_SHA points to for different events.

```suggestion
    - uses: actions/checkout@v2
      with:
        fetch-depth: 0
```

.github/workflows/publish.yml
```diff
@@ -20,12 +22,16 @@ jobs:
...
-        tag: ${{ github.ref }}
-        overwrite: true
-
+        token: ${{ secrets.GITHUB_TOKEN }}
+        allowUpdates: true
+        body: ${{ steps.tag_data.outputs.git-tag-annotation }}
+        artifact: archive/${{ steps.build_archive.outputs.archive }}
+        omitBodyDuringUpdate: true
+        omitNameDuringUpdate: true
+        # replacesArtifacts: true
```
RealOrangeOne, line 37: Comment

.github/workflows/publish.yml
WillB97: Yes, otherwise tags with `/` in them will misteriously not run

sedders123 (approved):

RealOrangeOne (approved):

PR #328, opened 2021-03-10
Use End State Scoring

sedders123:
Make it so the compstate uses end state scoring for new matches


modules/controller_utils/__init__.py
```diff
@@ -93,6 +93,7 @@ def record_arena_data(other_data: Dict[str, List[object]]) -> None:
     data = json.loads(match_file.read_text())
     arena_zones = data.setdefault('arena_zones', {})
     arena_zones['other'] = other_data
```
RealOrangeOne, line 95: Given we're mutating this now, we should probably copy it to be safe.

```suggestion
    arena_zones['other'] = other_data.copy()
```

WillB97 (approved):

PR #330, opened 2021-03-19
Lint whitespace in proto files

WillB97:
Fixes whitespace in the proto files as well as some of the long lines.

The whitespace fixes are done by a new script `proto-lint`.

This partially superseeds #318.


script/proto-lint
```diff
@@ -0,0 +1,99 @@
...
+    parser.add_argument(
+        '--quiet',
+        help="Suppress printing differing lines",
+        dest='print_diffs',
+        action='store_false',
+    )
+    return parser.parse_args()
+
+
+def main() -> int:
```
RealOrangeOne, line 87: Whilst technically correct, really this is returning `bool`. I'd recommend explicitly returning the error code rather than relying on bool casting.

WillB97: 0a524ec9e50dfdc3eb97e781806548ea32ba8b79

script/proto-lint
```diff
@@ -0,0 +1,101 @@
...
+            new_line = line.strip()
+
+            if re.search(r'^[})\]]', new_line):
+                current_indent -= 1
+
+            if current_indent < 0:
+                logging.error('Brackets mismatch, indentation went negative. Aborting')
+                return False
+
+            if new_line != "":
```
PeterJCLaw, line 37: nit
```suggestion
            if new_line:
```

PeterJCLaw (approved):
Checked the wrapping by multi-cursor editing back to the old format.

PR #333, opened 2021-03-21
Improve keyboard robot

WillB97:
- Add support for moving and turning together
- Print touch sensors as well as distance sensors
- Use async API for claims so you can move while claiming
- Add key for doing a radio sweep

### DIFF
--- a/controllers/example_controller/keyboard_controller.py
+++ b/controllers/example_controller/keyboard_controller.py
@@ -19,7 +19,7 @@ CONTROLS = {
 }
 
 
-def print_distance_sensors(robot: Robot) -> None:
+def print_sensors(robot: Robot) -> None:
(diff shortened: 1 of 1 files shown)


controllers/example_controller/keyboard_controller.py
```diff
@@ -1,7 +1,7 @@
 from sr.robot import Robot
 from controller import Keyboard
 
-KEYBOARD_SAMPLING_FREQUENCY = 16
+KEYBOARD_SAMPLING_PERIOD = 100
```
PeterJCLaw, line 4: Maybe worth a comment about how to choose this value now that it's substantially different than the timestep?

PeterJCLaw (approved):

PR #334, opened 2021-03-24
Improved territory_controller tests

WillB97:
- Adds tests to catch mismatched StationCode enums between the `territory_controller` and `sr.robot.radio`
- Makes tests less arena design dependent so changing the arena doesn't require changing the tests


controllers/territory_controller/tests.py
```diff
@@ -18,29 +20,81 @@
...
+        MockStationCode.T7,
     }
-    _zone_1_territories = {StationCode.PN, StationCode.EY, StationCode.PO, StationCode.YL}
-    _zone_1_disconnected = {StationCode.PN, StationCode.EY}
 
     def load_territory_owners(self, claim_log: ClaimLog) -> None:
-        # territories BG, TS, OX, VB, etc. owned by zone 0
         for territory in self._zone_0_territories:
-            claim_log._station_statuses[territory] = Claimant.ZONE_0
+            claim_log._station_statuses[territory] = Claimant.ZONE_0  # type: ignore
```
PeterJCLaw, line 76: Please could we have an [error code](https://mypy.readthedocs.io/en/latest/error_code_list.html) on this ignore?
I'm guessing that this is due to the use of a different set of station codes?
Should we perhaps make the `ClaimLog` generic so that it can handle different maps?

controllers/territory_controller/tests.py
```diff
@@ -179,18 +224,33 @@ def test_omitted_start_zones(self) -> None:
...
-                StationCode.BG,
-                StationCode.EY,
-                StationCode.OX,
-                StationCode.BE,
-                StationCode.PL,
-            },
-            'Territory VB has incorrect territory links',
+            station_codes,
+            radio_station_codes,
+            "StationCode enums differ between territory_controlled and sr.robot.radio",
```
PeterJCLaw, line 236: typo?
```suggestion
            "StationCode enums differ between territory_controller and sr.robot.radio",
```

controllers/territory_controller/tests.py
WillB97: af3ffdb3961af54e82c4f83ed78e3f814b653f55

sedders123 (approved):

PR #335, opened 2021-04-16
Arena Lighting Effects

WillB97:
Adds beginning and ending lighting. The lighting effects are configured using a cue stack that can support fading other light sources. Doesn't support running lighting effects before the start of the match although it has been noted that these could be recorded as a separate match with a custom cuestack and then stitched together afterwards.

https://user-images.githubusercontent.com/10937272/115083438-423bda80-9eff-11eb-952e-f4f3ff29404c.mp4

Fixes srobo/competition-simulator#17

PR #336, opened 2021-04-19
Make the compass' jitter magnitude independent of its value

WillB97:
Additionally allows the value to wrap around at the limits.

Currently the jitter is set to ±1.8°.

### DIFF
--- a/modules/sr/robot/compass.py
+++ b/modules/sr/robot/compass.py
@@ -16,4 +16,4 @@ class Compass:
         """
         x, _, z = self._compass.getValues()
         heading = atan2(x, z) % tau
-        return add_independent_jitter(heading, 0, tau, 0.5, True)
+        return add_independent_jitter(heading, 0, tau, 0.4, True)
--- a/modules/sr/robot/compass.py
+++ b/modules/sr/robot/compass.py
@@ -2,7 +2,7 @@ from math import tau, atan2
 
 from controller import Robot, Compass as WebotsCompass
 from sr.robot.utils import get_robot_device
-from sr.robot.randomizer import add_jitter
+from sr.robot.randomizer import add_independent_jitter
 
 
 class Compass:
@@ -16,4 +16,4 @@ class Compass:
         """
         x, _, z = self._compass.getValues()
         heading = atan2(x, z) % tau
-        return add_jitter(heading, 0, tau)
+        return add_independent_jitter(heading, 0, tau, 0.5, True)
--- a/modules/sr/robot/compass.py
+++ b/modules/sr/robot/compass.py
@@ -16,4 +16,4 @@ class Compass:
         """
         x, _, z = self._compass.getValues()
         heading = atan2(x, z) % tau
-        return add_independent_jitter(heading, 0, tau, 0.4, True)
+        return add_independent_jitter(heading, 0, tau, std_dev_percent=0.4, can_wrap=True)
--- a/modules/sr/robot/randomizer.py
+++ b/modules/sr/robot/randomizer.py
@@ -11,6 +11,9 @@ T = TypeVar('T', float, int)
 # The maximum randomness which can be added in either direction
 DEFAULT_RANDOM_RANGE_PERCENT = 2.5
 
+# The maximum randomness which can be added in either direction, compared to full scale value
+DEFAULT_RANDOM_INDEPENDENT_RANGE_PERCENT = 0.25
+
 
 def add_jitter(
     actual_value: T,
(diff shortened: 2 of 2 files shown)


prophile:
Code looks good, although practically I think it'd be better to draw from a gaussian distribution than a uniform one and set the standard deviation rather than the absolute range. Python has `random.gauss` for this so hopefully that wouldn't be much more code. Opinions?

I had a quick look at a ~£1 magnetometer on Farnell (LIS3MDLTR) which has a measurement error of stdev 0.365°. Throwing in the environmental factors on top, standard deviation 1.5° seems reasonable.

WillB97:
I agree, for sensor noise it really should be gaussian. Although we'd need to check that the reasonalby common outliers aren't that far off so teams that aren't using filtering don't end up with completely incorrect behaviour.

prophile:
With ~50 matches remaining, 2 teams per match, 120s each, sampling the compass 20Hz, that's about 240000 samples. With 1.5° standard deviation, the worst error we would expect to see in any single sample is 6.685°, which I think is OK.

WillB97:
Jitter is now gaussian distributed with σ=1.44°

prophile (approved):

modules/sr/robot/randomizer.py
```diff
@@ -21,3 +24,21 @@ def add_jitter(
...
+def add_independent_jitter(
+    actual_value: T,
+    min_possible: T,
+    max_possible: T,
+    std_dev_percent: float =  # % of full scale value
+    DEFAULT_RANDOM_INDEPENDENT_RANGE_PERCENT,
+    can_wrap: bool = False,
+) -> T:
+    value_range = max_possible - min_possible
+    random_range = value_range * (std_dev_percent / float(100))
```
RealOrangeOne, line 38: I wonder if the difference should be fixed instead? ie taking as an argument the +/- value explicitly.

modules/sr/robot/randomizer.py
```diff
@@ -21,3 +24,21 @@ def add_jitter(
...
     new_value = actual_value + random.uniform(-random_range, random_range)
     return type(actual_value)(max(min_possible, min(new_value, max_possible)))
+
+
+def add_independent_jitter(
+    actual_value: T,
+    min_possible: T,
+    max_possible: T,
+    std_dev_percent: float =  # % of full scale value
+    DEFAULT_RANDOM_INDEPENDENT_RANGE_PERCENT,
```
RealOrangeOne, line 34: Should this be on the same line? Reads weirdly this way

```suggestion
    std_dev_percent: float = DEFAULT_RANDOM_INDEPENDENT_RANGE_PERCENT,  # % of full scale value
```

modules/sr/robot/randomizer.py
WillB97: We could, although we'd want to remove the default value in that case.

modules/sr/robot/randomizer.py
WillB97: It was a line length issue but the current label fits (just).

modules/sr/robot/compass.py
```diff
@@ -16,4 +16,4 @@ def get_heading(self) -> float:
         """
         x, _, z = self._compass.getValues()
         heading = atan2(x, z) % tau
-        return add_jitter(heading, 0, tau)
+        return add_independent_jitter(heading, 0, tau, 0.4, True)
```
RealOrangeOne, line 19: 0.4 feels like a magic number here, where does that come from? If it's easy to calculate the stddev from a range, it could solve https://github.com/srobo/competition-simulator/pull/336#discussion_r615959663, too.

modules/sr/robot/randomizer.py
RealOrangeOne: I think that's ok, forces us to think about it. It also removes the magic number currently being passed, which is a little terse

WillB97: Essentially we'd be changing from a value relative to full scale to an absolute value. The upside of dealing in relative values is that it is a unit-less quantity so we don't need to bother with converting our constant between degrees and radians.

WillB97:
Note: this branch will not be merged before League 4 so that we can merge and use #335 in the league.

modules/sr/robot/compass.py
PeterJCLaw, line 19: Suggest passing `0.4` and `True` as kwargs for clarity.
```suggestion
        return add_independent_jitter(heading, 0, tau, std_dev_percent=0.4, can_wrap=True)
```

RealOrangeOne (approved):

PR #337, opened 2021-04-19
Better territory claim animations

WillB97:
Following the friendlies several improvements were suggested, this solves:
- LEDs are only accessed now if they would be changed
- Claims on cut-off towers don't light the top LED
- The tree of connected territories is only updated when it would have changed
  - This makes tests that regularly need whether a territory is claimable much cheaper

Currently the behaviour for a cut-off claim is to not light the top LED. While lighting up a different colour would be clearer the LED colour is additive to the default tower colour. This makes pure red LEDs almost indistinguishable while any other colour will have almost complete red value.

It is recommended to review each commit's diff separately.

### DIFF
--- a/controllers/territory_controller/territory_controller.py
+++ b/controllers/territory_controller/territory_controller.py
@@ -294,13 +294,14 @@ class ActionTimer:
     def __init__(
         self,
         action_duration: float,
-        progress_callback: Callable[[StationCode, Claimant, float], None] =
-        lambda station_code, claimant, progress: None,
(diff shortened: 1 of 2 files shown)


PeterJCLaw:
> * LEDs are only accessed now if they would be changed
> * The tree of connected territories is only updated when it would have changed

What does these mean? Are they just internal optimisations or are there an impact on the animations from them?

prophile (approved):

PR #339, opened 2021-08-18
Remove the ball joints in the crane

Adimote:
This makes the crane less wobbly, which is bad, but it does fix some serious issues with the hook after it has let go of the cube.


Adimote:
oops wrong repo!

PR #340, opened 2021-11-07
Updates for SR2022 microgames

WillB97:
Update the API to be inline with sr-robot3. This just provide a minimal empty arena so the simulator can be used with microgames to move around and get used to the API.


WillB97:
I know about the issue in the typing tests. I don't want to put type hints into the example code so I need a way to just ignore typing in that file.

trickeydan:
I haven't got time to do a full review I'm afraid. Looks like great work though, especially as we weren't planning to ship simulation this year. 🎉

A lot of the boilerplate code could be replaced by pulling in j5 as a dependency. I'm happy to do that after the physical kit is sorted out, but need to prioritise at the moment.

modules/sr/robot3/robot.py
```diff
@@ -36,14 +36,10 @@ def __init__(self, quiet: bool = False, init: bool = True) -> None:
         # thread, but not both.
         self._step_lock = Lock()
 
-        if init:
-            self.init()
+        self.init()
+        if not auto_start:
```
RealOrangeOne, line 40: I think this is the wrong way around

controllers/territory_controller/tests.py
```diff
@@ -233,9 +234,10 @@ def test_radio_matches_territory_controller(self) -> None:
         self.assertEqual(
             station_codes,
             radio_station_codes,
-            "StationCode enums differ between territory_controller and sr.robot.radio",
+            "StationCode enums differ between territory_controller and sr.robot3.radio",
         )
 
+    @unittest.skip("Territory controller no used with current arena")
```
RealOrangeOne, line 240: Nit
```suggestion
    @unittest.skip("Territory controller not used with current arena")
```

modules/sr/robot3/power.py
```diff
@@ -0,0 +1,109 @@
...
+        return 12.0
+
+    @property
+    def current(self) -> float:
+        return 0.0
+
+
+class Piezo:
+    def buzz(self, duration: float, note: 'Pitch') -> None:
+        pass
```
RealOrangeOne, line 83:
```suggestion
        print("🐝")  # noqa:T001
```

RealOrangeOne (commented):
I notice the documentation for the simulator has been removed from the docs. This'll need restoring if we ship this.

WillB97:
> A lot of the boilerplate code could be replaced by pulling in j5 as a dependency.

This was my feeling but this was just a quick fix to get us through kickstart and microgames.

RealOrangeOne (approved):
I haven't done a full API comparison, but from a skim it looks alright

modules/sr/robot3/power.py
WillB97: Lol no

RealOrangeOne (approved):

PR #347, opened 2022-10-01
Validate & update the supported libraries

PeterJCLaw:
This updates the supported libraries to the versions that it looks will be included in the physical kits. We also drop a couple which aren't going to be on the physical kits.

If teams want those dropped libraries then we'll need to tell them how to build them for their robot, so we can revisit whether we want them in the simulator at that point too.

### DIFF
--- a/libraries.txt
+++ b/libraries.txt
@@ -8,8 +8,6 @@
 # matches, so we *don't* have the full range of requirements.txt supported
 # syntax available.
 
-matplotlib==3.3.3
-numpy==1.19.3
-pandas==1.1.4
-scikit-learn==0.23.2
-scipy==1.5.4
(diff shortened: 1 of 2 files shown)


libraries.txt
```diff
@@ -8,8 +8,6 @@
...
 # syntax available.
 
-matplotlib==3.3.3
-numpy==1.19.3
-pandas==1.1.4
-scikit-learn==0.23.2
-scipy==1.5.4
+matplotlib==3.5.1
+numpy==1.22.3
+pandas==1.4.2
```
prophile, line 13: It is a real shame not to have these libraries.

prophile (approved):

PR #342, opened 2022-09-29
Update simulation files for Webots R2022b

PeterJCLaw:
There are a number of changes here, mostly automatically applied and broken into their own commits:
- update target Python & Webots versions in CI
- update the world & protos for R2022a (https://github.com/cyberbotics/webots/wiki/How-to-adapt-your-world-or-PROTO-to-Webots-R2022a)
- update the world for R2022b (https://github.com/cyberbotics/webots/wiki/How-to-adapt-your-world-or-PROTO-to-Webots-R2022b)

In the latter two cases the results here are a mix of scripted changes and minor manual fixes afterwards. See the individual commits for details.

The result is that this gets back to a working state for one robot and a Tin Can.
I'm deliberately not making changes to the actual arena here -- this is just about getting up to date with Webots.

Output from the CI run suggests that this is working ok: https://github.com/srobo/competition-simulator/suites/8569086174/artifacts/383263854
The mp4 recording seems fine, though the web animation shows a blank screen unless you click & drag slightly. Not sure if the latter failure was present previously; certainly it doesn't feel urgent right now.

PR #349, opened 2022-10-01
Drop Python 3.7, move minimum to 3.8

PeterJCLaw:
Per decision in Slack:
https://studentrobotics.slack.com/archives/C01CU87V94P/p1664643863983629?thread_ts=1664641945.025519&cid=C01CU87V94P

We'll actually be running on 3.10, however I'm leaving that change to when we move the webots version (#342) since the version of webots spec'd here doesn't support 3.10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ In order to run competition matches you'll need to:
    Note: you can change the version of Python which Webots uses from the UI --
    go to "Tools" > "Preferences" > "General" > "Python command".
 
-   We are using Python 3.7, though it shouldn't matter whether it's a system
+   We are using Python 3.8, though it shouldn't matter whether it's a system
    install or a virtual environment.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -145,7 +145,7 @@ In order to run competition matches you'll need to:
    Note: you can change the version of Python which Webots uses from the UI --
    go to "Tools" > "Preferences" > "General" > "Python command".
 
-   We are using Python 3.7, though it shouldn't matter whether it's a system
+   We are using Python 3.8, though it shouldn't matter whether it's a system
```
prophile, line 148: What does this paragraph mean bereft of the virtual competition context?

PeterJCLaw: Not much, though I don't think this is the PR to address that.

prophile (approved):

Issue #199, opened 2020-10-24
Compatibility with Kit API


trickeydan:
This simulator currently exposes a Python API that is not compatible with our existing Python 2 kit, nor the upcoming [sr-robot3](https://github.com/srobo/sr-robot3).

Whilst the new sr-robot3 API is now delayed until SR2022 due to SR2021 being virtual, this is an issue that needs to be considered. I recall [pre-empting this issue](https://github.com/srobo/competition-simulator/issues/5#issuecomment-626163844) when this simulator was designed, although this was [subsequently ignored](https://github.com/srobo/competition-simulator/pull/32#issuecomment-628200440) in a PR that was merged within **3 minutes** of creation (no chance for review!).

Causing breaking changes to the API too frequently will confuse both students and volunteers, which is why we planned to originally limit the breaking changes to coinciding with the move to Python 3 and sr-robot3.

There is no ideal solution to this situation, and we can only learn from mistakes that have been made here. I suspect that the best solution to this is to make use of the functionality for multiple backends that is built into sr-robot3, and ship that as the API for SR2021.

trickeydan:
It looks like it shouldn't be too difficult to get this working with sr-robot3.

## Requirements

- Add sr-robot3 as a dependency of the competition simulator
- Subclass sr.robot3.Robot to add webots specific functions
- Implement simulator backends (see [https://github.com/srobo/competition-simulator/pull/32#issuecomment-628200440](https://github.com/srobo/competition-simulator/pull/32#issuecomment-628200440)
    - SRV4PowerBoardSimulatedBackend
    - SRV4MotorBoardSimulatedBackend
    - SRV4ServoBoardSimulatedBackend
    - SRV4CameraSimulatedBackend
    - SRV4RuggeduinoSimulatedBackend

Some additional work will likely be required to get vision working, but all of the hard work has been done for it already.

PeterJCLaw:
Are there substantial ways that this API breaks from the existing (Python 2) kit API other than due to missing/added features or by merely being Python 3 rather than Python 2? If there are concrete things which can be fixed then we can consider fixing that for SR2021.

We should also bear in mind that we now have a non-trivial number of volunteers who have experience with the API which this simulator exposes.

As an aside: we are also discussing whether or not we need to provide an API wrapper at all given that Webots' API is already in Python and that encouraging the teams to use it would allow them much more flexibility in the robot chassis that they enter.

trickeydan:
[...]

PR #350, opened 2022-10-16
Update the simulated robot API to (mostly) match sr.robot3

PeterJCLaw:
This updates the simulated `sr.robot3` API to mostly match the one in the published package of the same name.

There are some differences here, mostly driven by what the simulator doesn't (usually cannot) support. In general I've gone with completely omitting members which don't work in the simulator, though in a few places I've implemented dummy versions that record the nature of the simulator.

The split here is approximately between things that it seems "safe" to emulate as a no-op versus those which are likely to change the behaviour of the robot substantially. Many of these are fuzzy, so we might want to revisit the decisions, but it probably makes sense to get these changes live first and iterate.

The changes here were determined through a mixture of code inspection of the `sr.robot3` source and comparison against the docs. The latter was a mix of manual comparison and using https://gist.github.com/PeterJCLaw/318d392c46daff36577df5e24866d5b0 to gather sample code to run `mypy` against.

Not included here are:
- KCH LEDs: I'd like to have these in, however as their API is in flux (https://github.com/srobo/sr-robot3/pull/66) and I'd ideally like to actually implement the LEDs I plan to do this in a separate PR.
- Vision: this is going to be rather complicated because Zoloto is a) very different than `libkoki` and b) a bit of a mess right now (https://github.com/RealOrangeOne/zoloto/pull/302, https://github.com/RealOrangeOne/zoloto/pull/304). I plan to also pick this up as a separate PR.

If we have any example robot code which works for `sr.robot3` it would be handy to try it against this PR's simulator and see what works/what doesn't.

Fixes https://github.com/srobo/competition-simulator/issues/199


Adimote (approved):
LGTM!

PR #359, opened 2023-01-07
Add support for R2023a

WillB97:
And updates types so Webots R2023a can run.


RealOrangeOne:
Suggestion: We should update the CI matrices to run tests against R2023a

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -193,6 +193,14 @@ def get_simulation_run_mode(supervisor: Supervisor) -> SimulationMode:
         return Supervisor.SIMULATION_MODE_FAST
 
 
+def set_simulation_mode(supervisor: Supervisor, mode: SimulationMode) -> None:
+    try:
+        supervisor.simulationSetMode(mode)
+    except AttributeError:
```
sedders123, line 199:
```suggestion
    except AttributeError:
        # Webots R2023a throws an AttributeError so we'll call underlying setter ourselves
```

sedders123 (approved):

controllers/competition_supervisor/competition_supervisor.py
```diff
@@ -193,6 +193,15 @@ def get_simulation_run_mode(supervisor: Supervisor) -> SimulationMode:
...
+        # Webots R2023a throws an AttributeError so we'll call underlying setter ourselves
```
PeterJCLaw, line 200: This makes it sound like a bug -- is that expected? If so, should we report this upstream?

Separately I'd be tempted to guard _only_ the attribute lookup we're using as detection of the runtime version, so that any other `AttributeError`s don't cascade into hard-to-diagnose issues.

setup.cfg
```diff
@@ -77,3 +78,6 @@ warn_unused_configs = True
 # Our stubs are deliberately importable, rather than purely stubs,
 # meaning that mypy treats them as code rather than stubs.
 disable_error_code = empty-body
+
+[mypy-controller.device]
+disable_error_code = empty-body
```
PeterJCLaw, line 83: Suggest merging this with the above since they're logically the same:
```
[mypy-controller,controller.device]
```

stubs/controller/wb.py
```diff
@@ -0,0 +1,3 @@
+from typing import Any
+
+wb: Any  # type: ignore[misc]
```
PeterJCLaw, line 3: Please can we add some structure here rather than using `Any`?

controllers/competition_supervisor/competition_supervisor.py
WillB97: This bug has actually been fixed upstream and will be in the next release. https://github.com/cyberbotics/webots/blob/master/docs/reference/changelog-r2023.md?plain=1#L15

stubs/controller/wb.py
WillB97: Since we are only useing `wb` to work around a bug I've removed the stub.

modules/sr/robot3/camera.py
```diff
@@ -191,20 +192,21 @@ def _see(self) -> list[Marker]:
                 recognition_model,
             )
             if marker_info:
-                object_infos[recognition_object] = marker_info
+                recognition_objects.append(recognition_object)
+                object_infos[recognition_object.getModel()] = marker_info
```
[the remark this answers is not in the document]
WillB97: `getModel` just returns the contents of the model field which does not have a unique constraint. I think this can be done by using `getId` instead which returns the object's ID in the scene tree.

PeterJCLaw (approved):

PR #366, opened 2023-01-15
Approximately match kit's vision implementation

### LINKED ISSUE #360: Line up vision information with the kit API
Most notably the orientation information is different.

We should also review the other properties as the kit API's vision has changed.

WillB97:
Moves the vision implementation to use the output data structures used by the kit. Cartesian, Spherical and Orientation have been verified.

Outstanding:
- [x] ~Remove numpy dependency~ (Now uses [MomsFriendlyRobotCompany/squaternion](https://github.com/MomsFriendlyRobotCompany/squaternion))
- [x] Fix occlusion, the previous method using bounding boxes risks producing a lot of both false positives and negatives. Using additional recognition objects to test that the corners are visible should cover most cases and keeps most the occlusion calculations in opengl not python.
- [ ] For both performance improvements and to bring it more inline with the kit UI the camera view is only updated when an image detection is done.

Currently only tested in R2023a.

Fixes #360

Builds on #359


trickeydan:
> Remove numpy dependency (Could be solved by switching from pyquaternion to [MomsFriendlyRobotCompany/squaternion](https://github.com/MomsFriendlyRobotCompany/squaternion))

Please don't do this in the middle of the competition. That suggested library has not been tested in production, is very new and is not widely used. Switching to that library is a decision that would need to be agreed in a _kit team meeting_ please.

trickeydan:
What is the impact of changing the quaternion library on the rest of the kit?

How robust is this new library? It is not widely used, has limited tests and no typing.

Can we be sure that dropping numpy won't break student code? Is this definitely backwards compatible?

Why are we copying in the code rather than submoduling or using a PyPI package?

PeterJCLaw:
Some context here:

The simulator distribution doesn't (currently) have any dependencies during development. This is very much by design -- so that competitors don't need to install anything extra to use it. While that may be a decision to revisit in a future year, I don't think it's something we want to change during the competition year.

The removal of numpy here is presumably(?) a change relative to early versions of the PR rather than to what's released -- the simulator as published doesn't have any dependency on numpy (though competitor code is welcome to as we guarantee that it'll be present in the competition environment).

The same is true of the quaternion library -- previously there wasn't one, so if there's one here now and the API is compatible I think it's fine (on that front). (No comment on its suitability or quality)

WillB97:
Squaternion is a pure-python implementation of the quaternion maths we need, teams will have numpy in the virtual competition but may not have it during development.

The remaining differences from the kit API are:
- There is not a `R.camera.capture()` method.
- None of the camera methods support a `frame` argument as the recogniion system cannot be given arbitrary frames.
- There is not a `R.camera.save()` method.
- The quaternion returned by `Marker.orientation.quaternion` is only a tuple of the quaternion's w, x, y, z components.

Adimote (approved):
looks good to me, though my review wasn't incredibly in-depth due to the quantity

modules/sr/robot3/camera.py
```diff
@@ -1,229 +1,156 @@
...
-        raise AssertionError("Unknown object type")
+from .utils import maybe_get_robot_device
 
+LOGGER = logging.getLogger(__name__)
 
-MARKER_SIZES: dict[Container[int], int] = {
+MARKER_SIZES: dict[Iterable[int], int] = {
     range(28): 200,  # 0 - 27 for arena boundary
-    range(28, 100): 100,  # Everything else is a token
+    range(28, 100): 80,  # Everything else is a token
```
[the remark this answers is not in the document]
WillB97: Token markers are now 80mm in cd5b0a1b160c625a76ac20d28fef7cb96e2eab12

modules/squaternion/readme.md
```diff
@@ -0,0 +1,32 @@
+# Simple Quaternions (`squaternion`)
+Vendored from [MomsFriendlyRobotCompany/squaternion](https://github.com/MomsFriendlyRobotCompany/squaternion) on 2023-01-16.
```
PeterJCLaw, line 2: Could we include the commit hash for the vendored revision?

PeterJCLaw: https://github.com/srobo/competition-simulator/pull/366/commits/e373663966baeb19ce3ab8c0a9526037fd9aada9

modules/sr/robot3/camera.py
```diff
@@ -1,229 +1,190 @@
...
+        :returns: list of markers that the camera could see.
+        """
+        raw_markers: list[Marker] = []
+        recognition_objects: dict[str, dict[str, Recognition]] = defaultdict(dict)
+
+        with self._capture():
+            for recognition_object in self._camera.getRecognitionObjects():
+                # Get the object's assigned "model" value, the marker has 5 detection points:
+                # the marker itself and the 4 corners, named in the form <id>_<location>
+                tag_uid, tag_name, tag_point = str(recognition_object.getModel()).split('_')
```
PeterJCLaw, line 85: Note that because `getModel` returns `bytes` on R2022b, calling `str` on that gives a string like `"b'1234'"`. We should probably extract a function `force_text` which does the conversion without adding the bytes `repr` to the string.

protos/Markers/MarkerBase.proto
```diff
@@ -33,7 +118,7 @@ PROTO MarkerBase [
       }
     ]
     name IS name
-    model IS model
+    model %<= "\"" + uid + "_" + fields.model.value + "_base\"" >%
```
PeterJCLaw, line 121: This looks a lot like it will create unique identifiers for each marker, potentially providing competitors with a means (at least within a given match and for tokens they've already interacted with) to infer the kind of token from its marker. That seems undesirable given that not being able to do that was one of the core design principles of the game.

PeterJCLaw:
Testing this locally in R2022b using the keyboard robot fails to find any markers, even when the image overlay does show markers in shot which are annotated with the red border (i.e: the recognition objects have been found).

PeterJCLaw (changes requested):
This doesn't seem to work for me (in R2022b, with the keyboard robot, on Ubuntu 22.04).

modules/sr/robot3/camera.py
PeterJCLaw: 8181cdb fixes this, making marker recognition work on R2022b.

controllers/example_controller/keyboard_controller.py
```diff
@@ -71,10 +71,13 @@ def print_sensors(robot: Robot) -> None:
                 print(
                     f" Spherical: {angle_str(rot_x)}, {angle_str(rot_y)}, {dist}",
                 )
-                rot_x, rot_y, rot_z = marker.orientation
+                roll = marker.orientation.roll
+                pitch = marker.orientation.pitch
+                yaw = marker.orientation.yaw
```
PeterJCLaw, line 76: Presumably this is expanded due to the bug in april_vision? Can we not just fix the bug?
https://github.com/WillB97/april_vision/issues/7 should be fixed by https://github.com/WillB97/april_vision/pull/9.

PeterJCLaw:
The CI failure here is due to the tuple re-ordering as noted in https://github.com/WillB97/april_vision/issues/8. Since fixing that is potentially itself breaking the solution isn't clear right now, however I definitely don't think we should introduce the same breakage into the simulator.

controllers/example_controller/keyboard_controller.py
WillB97: No, this change was to make it clearer what the rotations are. Roll, pitch and yaw are understood by the layman and axis agnostic.

protos/Markers/MarkerBase.proto
WillB97: This is a unique value for each token so that the additional points can be linked to the correct token to emulate occlusion. At this level of the API the students could equally just call getNode to get the unique node identifier of every instance of a marker.

PeterJCLaw:
My testing of this seems to be not finding markers when they're near the edge of the camera's field of view, including cases where the maker is _only just_ overlapping the edge of the image; cases where I would definitely expect the real vision system to pick up the marker. Is that expected?

PeterJCLaw:
> My testing of this seems to be not finding markers when they're near the edge of the camera's field of view, including cases where the maker is _only just_ overlapping the edge of the image; cases where I would definitely expect the real vision system to pick up the marker. Is that expected?

https://github.com/srobo/competition-simulator/pull/366/commits/ec23ea7cb67c1033fa8f05f2735cac73465bd16a
fixes this by moving the corner points inwards so we're not caring about the whitespace boundary.

PeterJCLaw:
Testing suggests that this allows detection of markers at arbitrarily oblique angles, something that the previous implementation did not (it capped recognition at 75°). We should probably reinstate this to avoid silliness.

WillB97:
> Testing suggests that this allows detection of markers at arbitrarily oblique angles, something that the previous implementation did not (it capped recognition at 75°). We should probably reinstate this to avoid silliness.

The current kit vision is capable of detecting markers at oblique angles (beyond 90° at the edge of images). Given that they only have 3.5 weeks left I don't see an issue with vision being slightly better.

PeterJCLaw:
> The current kit vision is capable of detecting markers at oblique angles (beyond 90° at the edge of images).

No, it isn't. Beyond 90° is physically impossible (that implies the marker is facing away from the camera) and local testing suggests that the borderline cases where the marker is very oblique are also not picked up.

8BitJosh:
> > The current kit vision is capable of detecting markers at oblique angles (beyond 90° at the edge of images).
> 
> No, it isn't. Beyond 90° is physically impossible (that implies the marker is facing away from the camera) and local testing suggests that the borderline cases where the marker is very oblique are also not picked up.

image

Beyond 90 isnt impossible, the angle is to the flat plane perpendicular to the cameras view, not the angle relative to the cameras position. See the attached diagram as an example

PeterJCLaw:
> Beyond 90 isnt impossible, the angle is to the flat plane perpendicular to the cameras view, not the angle relative to the cameras position. See the attached diagram as an example

Ah, I think you've misunderstood what calculation I'm doing. I'm comparing the orientation to the vector along the line from the camera to the marker. Thus the zero angle (and the 90°) is relative to the position of the marker. 90° always represents a marker arranged perfectly edge-on to the camera.

PR #372, opened 2023-02-04
Add pressure-sensing fingers

Adimote:
Adds fingers that fold down and lift up the front of the robot, thus lifting up a cube. The ends of the fingers are pressure sensors, in Arduino analog pins A6 (notably these don't exist on real arduinos), which then return a number proportional to the weight of the robot plus the object being carried.

See Video Demo:
https://user-images.githubusercontent.com/2551763/216791604-d9f9adf9-2754-49e4-b0d4-ab22b4c83f3e.mp4

### DIFF
--- a/controllers/example_controller/keyboard_controller.py
+++ b/controllers/example_controller/keyboard_controller.py
@@ -17,6 +17,8 @@ CONTROLS = {
     "boost": (Keyboard.SHIFT, Keyboard.CONTROL),
     "grabber_open": (ord("R"), ord("P")),
     "grabber_close": (ord("E"), ord("O")),
+    "fingers_down": (ord("C"), ord(",")),
+    "fingers_up": (ord("X"), ord(".")),
     "angle_unit": (ord("B"), ord("B")),
 }
 
@@ -44,6 +46,9 @@ def print_sensors(robot: Robot) -> None:
     touch_sensor_names = [
         "Rear",
     ]
+    pressure_sensor_names = {
+        A6: "Fingers",
+    }
 
     print(f"Distance sensor readings at {robot.time():.2f}s:")
     for Apin, name in distance_sensor_names.items():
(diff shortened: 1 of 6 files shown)


controllers/example_controller/keyboard_controller.py
```diff
@@ -122,20 +134,20 @@ def print_sensors(robot: Robot) -> None:
...
-            left_power -= 0.25
-            right_power += 0.25
+            left_power -= 0.125
+            right_power += 0.125
 
         elif key_ascii == key_right:
-            left_power += 0.25
-            right_power -= 0.25
+            left_power += 0.125
+            right_power -= 0.125
```
PeterJCLaw, line 150: Please can we break this speed change out into a separate PR.

PeterJCLaw:
Numbers I'm seeing:
- 1.0 ish for just the robot
- 1.3 ish for Bronze
- 1.6 ish for Silver
- 1.9 ish for Gold

Given that we're expecting the Gold tokens to weigh about the same as a Silver, this may not be representative.

WillB97 (approved):

controllers/example_controller/keyboard_controller.py
Adimote: removed these changes

WillB97:
I get the error:
> WARNING: DEF ROBOT-0 Robot_2022 (PROTO) > DEF FINGERS Transform > DEF FINGER_HINGE HingeJoint > Solid > TouchSensor: Webots has detected that this solid is light and oblong according to its inertia matrix. This belongs in the physics edge cases, and can imply weird physical results. Increasing the weight of the object or reducing its eccentricity are recommended.

Adimote:
> I get the error:
> 
> > WARNING: DEF ROBOT-0 Robot_2022 (PROTO) > DEF FINGERS Transform > DEF FINGER_HINGE HingeJoint > Solid > TouchSensor: Webots has detected that this solid is light and oblong according to its inertia matrix. This belongs in the physics edge cases, and can imply weird physical results. Increasing the weight of the object or reducing its eccentricity are recommended.

now fixed, just had to add 0.05kg :)

Adimote:
I've now split the fingers into two separate servos/pressure sensors. It looks like they normally measure 0.5* the value when they were one, so students can just add them together to get the total pressure.

Also it means I no longer need to tips of the fingers to be 0.2kg, they're now neglegibly light, so the robot is 0.2kg lighter.

PeterJCLaw (approved):

PR #374, opened 2023-02-08
Vision clarifiations

PeterJCLaw:
Suggest review by commit

### DIFF
--- a/modules/sr/robot3/marker.py
+++ b/modules/sr/robot3/marker.py
@@ -216,18 +216,23 @@ class Orientation:
         angle=pi, axis=(1, 0, 0),
     )
 
-    def __init__(self, rotation_matrix: tuple[float, float, float, float]):
+    @classmethod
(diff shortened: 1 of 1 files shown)


modules/sr/robot3/marker.py
```diff
@@ -216,18 +216,23 @@ class Orientation:
...
     )
 
-    def __init__(self, rotation_matrix: tuple[float, float, float, float]):
+    @classmethod
+    def from_axis_angle(cls, axis_angle: tuple[float, float, float, float]) -> Orientation:
+        x, y, z, rad = axis_angle
+
+        # Calculate the quaternion of the rotation in the camera's coordinate system
+        # Correct for roll and yaw turning the wrong way
+        quaternion = Quaternion.from_angle_axis(angle=rad, axis=(-x, y, -z))
```
WillB97, line 225: Can we not move the axis conversion from the init as we want to keep the quaternion being passed in webots' coordinate system.

WillB97 (approved):

PR #376, opened 2023-02-12
Revert vision changes

PeterJCLaw:
The recent vision improvements are technically breaking, additionally they unfortunately still have the underlying bug where orientation data is messed up (due to [orientation issues in Webots](https://discord.com/channels/565154702715518986/1071170269080719430/1071170332276314162)),

They do however fix an issue where markers sometimes can't be seen (which is actually a result of the orientation data being messed up). If we're going to do breakage it would be nice to do it all in one go, once we've fixed the orientation data.

This PR reverts the recent vision changes so we can put out a release with the other changes separately.

Note: I haven't tested this at all -- it's just a revert of the commits I'm aware touched vision stuff, plus a small fix for CI.


Adimote:
I did a few checks of X/Y orientation on a marker in a few orientations, and cartesian distances, it seems like the measurements are the same as pre-vision changes (commit 64d5994) tag: sr2023.3

Adimote (approved):

PR #383, opened 2023-02-18
Add script to fix animations on windows

### LINKED ISSUE #378: run-comp-match animations are broken
Webots animation files now record the absolute path to their associated files so copying them from a temp directory causes them to not load correctly.

WillB97:
Makes absolute paths relative, assumes ancillary files are siblings of the html.

Fixes #378

### DIFF
--- a/script/fix-animation
+++ b/script/fix-animation
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+import re
+import argparse
+
+
+def fix_animation(input: str, output: str) -> None:
+    with open(input) as f:
+        data = f.readlines()
+
+    output_data = []
+    for line in data:
+        if 'stylesheet' in line:
+            match = re.search(r'href="(.*?)"', line)
+            if match is None:
+                output_data.append(line)
+            else:
+                css_name = match[1].replace('\\', '/').split('/')[-1]
+                output_data.append(f'    <link rel="stylesheet" href="{css_name}">\n')
+        elif '<webots-view' in line:
+            match = re.search(
+                r'data-thumbnail=(.*?) .*data-scene=(.*?) .*data-animation=(.*?)>', line)
+            if match is None:
+                output_data.append(line)
+            else:
+                jpg_name = match[1].replace('\\', '/').split('/')[-1]
+                x3d_name = match[2].replace('\\', '/').split('/')[-1]
+                json_name = match[3].replace('\\', '/').split('/')[-1]
+                output_data.append(
+                    f'      <webots-view data-thumbnail={jpg_name} '
+                    f'data-scene={x3d_name} data-animation={json_name}></webots-view>\n',
+                )
+        else:
+            output_data.append(line)
+
+    with open(output, 'w') as f:
+        f.writelines(output_data)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('animation')
+    parser.add_argument('output')
+
+    args = parser.parse_args()
+
+    fix_animation(args.animation, args.output)
+
+
+if __name__ == '__main__':
+    main()


script/fix-animation
```diff
@@ -0,0 +1,51 @@
...
+        if 'stylesheet' in line:
+            match = re.search(r'href="(.*?)"', line)
+            if match is None:
+                output_data.append(line)
+            else:
+                css_name = match[1].replace('\\', '/').split('/')[-1]
+                output_data.append(f'    <link rel="stylesheet" href="{css_name}">\n')
+        elif '<webots-view' in line:
+            match = re.search(
+                r'data-thumbnail=(.*?) .*data-scene=(.*?) .*data-animation=(.*?)>', line)
```
PeterJCLaw, line 21: nit: please could we wrap this for consistency:
```suggestion
                r'data-thumbnail=(.*?) .*data-scene=(.*?) .*data-animation=(.*?)>',
                line,
            )
```

script/fix-animation
```diff
@@ -0,0 +1,51 @@
...
+
+    output_data = []
+    for line in data:
+        if 'stylesheet' in line:
+            match = re.search(r'href="(.*?)"', line)
+            if match is None:
+                output_data.append(line)
+            else:
+                css_name = match[1].replace('\\', '/').split('/')[-1]
+                output_data.append(f'    <link rel="stylesheet" href="{css_name}">\n')
```
PeterJCLaw, line 18: Feels like it would be safer to do a find & replace on the adjusted text rather than fully rebuilding the line. That would avoid any issues with the underlying template changing (also below).

script/fix-animation
```diff
@@ -0,0 +1,51 @@
...
+        elif '<webots-view' in line:
+            match = re.search(
+                r'data-thumbnail=(.*?) .*data-scene=(.*?) .*data-animation=(.*?)>', line)
+            if match is None:
+                output_data.append(line)
+            else:
+                jpg_name = match[1].replace('\\', '/').split('/')[-1]
+                x3d_name = match[2].replace('\\', '/').split('/')[-1]
+                json_name = match[3].replace('\\', '/').split('/')[-1]
```
PeterJCLaw, line 27: Maybe worth asserting that the removed portion of these is all the same?
Perhaps like:
```suggestion
                *jpg_dir, jpg_name = match[1].replace('\\', '/').split('/')
                *x3d_dir, x3d_name = match[2].replace('\\', '/').split('/')
                *json_dir, json_name = match[3].replace('\\', '/').split('/')
                assert jpg_dir == x3d_dir == json_dir
```

script/fix-animation
```diff
@@ -0,0 +1,51 @@
...
+
+    with open(output, 'w') as f:
+        f.writelines(output_data)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('animation')
+    parser.add_argument('output')
```
PeterJCLaw, line 43: For learnings: you might be interested in [`argparse.FileType`](https://docs.python.org/3/library/argparse.html#argparse.FileType), a class which can be passed to the `type` kwarg here. It handles opening the files as well as natively supporting using `-` for STDIN and or STDOUT.

script/fix-animation
```diff
@@ -0,0 +1,51 @@
...
+import re
+import argparse
+
+
+def fix_animation(input: str, output: str) -> None:
+    with open(input) as f:
+        data = f.readlines()
+
+    output_data = []
+    for line in data:
```
PeterJCLaw, line 11: General style point for learning: I would encourage thinking about separating the logic which processes the content of the files from the logic which handles the file I/O. Doing so tends to make the code clearer as well as making the processing logic more easily reusable.
For this case I'd probably go with a separate function which operates on a single line at a time (and have a comment explicitly noting the assumption that the HTML elements of interest are each on a single line), keeping the loop in the same function as the IO.

PeterJCLaw (approved):
Nice, thanks for putting this together. Some small suggestions inline to make this a bit more robust.

PR #384, opened 2023-02-18
Save an image at the end of the match

WillB97:
Sorts out lighting for the image outside of the video.
match-4

https://user-images.githubusercontent.com/10937272/219864202-dfa0e357-9350-498a-9f9b-6dd90aefc82b.mp4


PeterJCLaw:
Nice! Would be great to have some comments about how/why this works. I realise we don't quite understand ourselves, however I suspect that means it'll be even harder to understand in future.
I think the important bits to capture are:
- cross link comments between the image capture and the lighting, so that the dependency is clear
- noting the nuances of the timing which are needed to get the behaviour we want out of Webots.

PeterJCLaw (approved):

PR #388, opened 2023-02-22
Set textures for tokens to increase visibility

Adimote:
Add 1px textures for Gold, Silver, and Bronze tokens so they are easier to discern.

This has quite a lot of repetition in the token proto files, sadly I couldn't figure out how to DEF a string list.

Before:
image

After:
image


PeterJCLaw:
Bear in mind that textures don't carry over into the animations. Not sure if that's a big issue though?

PeterJCLaw:
Thought -- does this need to be texture, or could we make the marker a specific colour?

PeterJCLaw:
Closing in favour of #389.

PeterJCLaw:
Closing as this was specific to last year's game. We might revisit this idea for future tokens, however it will likely want different colours/etc.

PR #389, opened 2023-02-22
Set marker visual size to 80mm, retain actual size

Adimote:
Add a visual_size field to MarkerBase, so the visual size can be different from the actual size (used by the camera).


PeterJCLaw:
Have tested that this doesn't impact vision by launching the keyboard robot and pressing Q without moving the robot, then diffing the console output vs on `main` -- no changes to vision output (at the 2sf we have for vision properties).

PeterJCLaw (approved):

WillB97 (approved):

PR #390, opened 2023-02-24
Move marker visuals 2mm forwards to prevent z-fighting

### LINKED ISSUE #387: z-fighting on token markers
There's some z-fighting visible for token markers if they're on the far end of the arena, see screenshot:
image

Adimote:
Move the visual element of a marker forwards by 2mm to prevent z-fighting in the recordings. This does not change the detections of the markers at all

Before:

https://user-images.githubusercontent.com/2551763/221275105-136ec474-3d69-453c-9c94-1268e49b7b92.mp4

After:

https://user-images.githubusercontent.com/2551763/221275121-28817437-dbb5-4c9c-9e70-26adc22e5ef5.mp4

fixes #387

### DIFF
--- a/protos/Markers/MarkerBase.proto
+++ b/protos/Markers/MarkerBase.proto
@@ -18,19 +18,24 @@ PROTO MarkerBase [
     translation IS translation
     rotation IS rotation
     children [
-      Shape {
-        appearance PBRAppearance {
-          baseColorMap ImageTexture {
-            url IS texture_url
-            repeatS FALSE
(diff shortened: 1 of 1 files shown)


PeterJCLaw (approved):

PeterJCLaw:
@Adimote do you think we should merge this? Clearly we managed without it, so the issue can't have been too bad.

Separately: would a 1mm shift be sufficient? Feels like that would be visually preferable if so.

PR #395, opened 2023-09-02
Test our code on Python 3.11

PeterJCLaw:
Even though the versions of Webots we're using don't yet support it, we expect that that will change as we move to newer versions.

Builds on #400 for cleaner CI.


Adimote (approved):
Looks good to me, though we should probably remove the 3.8 tests?

PR #400, opened 2023-09-02
Update our promised libraries to recent versions

PeterJCLaw:
We don't know yet what versions will be on the kits, however these updates are needed to unblock Python 3.11 support so it's likely that similar updates will happen there. Once we know what versions the kits will support this year we can update these again easily.

See also https://github.com/srobo/docs/issues/496

Builds on #396 for cleaner CI.

### DIFF
--- a/libraries.txt
+++ b/libraries.txt
@@ -19,6 +19,7 @@ shapely==2.0.1
 scipy==1.11.2
 scikit-learn==1.3.0
 networkx==3.1
+# Note: deliberately no 'flask' here as it's unlikely to be useful in the simulator
 
 # An extra offered in the simulator
 typing-extensions==4.7.1
--- a/libraries.txt
+++ b/libraries.txt
@@ -16,7 +16,7 @@ pandas==2.0.3
 numpy==1.24.4
 pillow==10.0.0
 shapely==2.0.1
-scipy==1.11.2
+scipy==1.10.1
 scikit-learn==1.3.0
 networkx==3.1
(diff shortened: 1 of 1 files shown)


libraries.txt
```diff
@@ -8,7 +8,21 @@
...
+# List from https://github.com/srobo/robot-image/blob/main/files/python/libraries.txt
+matplotlib==3.7.2
+# Note: pandas 2.1 has some known issues, avoid it for now.
+# (https://github.com/pandas-dev/pandas/issues?q=is%3Aissue+is%3Aopen+label%3ARegression)
+pandas==2.0.3
+numpy==1.24.4
+pillow==10.0.0
+shapely==2.0.1
+scipy==1.11.2
+scikit-learn==1.3.0
```
Adimote, line 20: I notice flask is in https://github.com/srobo/robot-image/blob/main/files/python/libraries.txt but omitted here, is that intentional (I'm not sure why they'd need flask anyway, to be honest)

Adimote (approved):

libraries.txt
PeterJCLaw: I think flask is potentially useful on the robots if teams want to integrate with/provider their own web UI. I don't think it's useful in the simulator, so it's intentional to drop it. I should have added a comment about that.
https://github.com/srobo/competition-simulator/pull/400/commits/b2f61cfd9ab47da3abbc6caca85ef6ebec19739c

PR #415, opened 2023-09-23
SR2024 vision API and full stack integration test harness

### LINKED ISSUE #377: Update vision API to match SR2024 kit
Alternatively we need to carefully document exactly what does and doesn't work.

PeterJCLaw:
This updates the vision API to match that expected by the SR2024 kit.

## Changes

Included here:
- position and orientation processing updates
- zeroing the basis of markers to be aligned to Webots axes, making unit tests data considerably easier to manually check
- a full-up integration test harness which will make future work in this area _much_ easier to validate
- a script for generating complex (multi-axis) poses & matching tests; along with validation that it is in sync with its outputs

Not included here:
- addressing marker sizing (#417) as this will require adjusting the tokens/etc. and this PR is already huge
- complete tidyup of the legacy code; I've remove stuff which was obviously dead, but haven't chased down everything that's no longer used. Note that some of the legacy style code is still used as part of the tests.
- the check we used to have that markers weren't at too oblique an angle. This is added back in https://github.com/srobo/competition-simulator/pull/420, which I'd like to merge into this PR once both are ready (i.e: and them merge to `main` together).

## Code review

I suggest we focus on the correctness of the data returned from the API and whether the API shape matches the kit API. It may also be worth splitting the review as I realise the diff is pretty big.

I'm happy to do a walk-through review if that would be useful, alternatively here's a guide to the changes.

The core things to review (and I suggest this order) are:
- the new tests (`controllers/test_supervisor/test_supervisor.py`), including validating that the manually constructed test cases there are correct (but don't worry about the generated ones yet)
  - related world files:
    - `protos/Components/SRCamera.proto`
    - `worlds/Tests.wbt` (again ignore the generated cases for now)
  - related plumbing: `script/testing/integration-test`, `.github/workflows/integration.yml` (renamed from `run-match.yml`)
- `modules/sr/robot3/camera.py` (it may be easier to review this as if it were new rather than a diff)
  - the actual computations happen in `modules/sr/robot3/vision/convert.py` and `modules/sr/robot3/vision/markers.py` (renamed and simplified from `tokens.py`, reviewing the diff of that may be helpful)
  - various new types are copied over either from the kit API or `april_vision`; I've tried to comment these inline so they're easy to keep in sync

That covers the core of the logic. Much of the rest of the changes are accounting for the change in alignment of the marker proto to webots axes or are other plumbing:
- `*.proto` and `Arena.wbt` files shouldn't have changed in a way which changes the SR2023 arena -- that should still look exactly as it did and work the same
- `script/testing/insert-poses.py` computes the possible combinations of π/4 rotations on each axis and inserts them reproducibly into both our testing world file and our test supervisor; there's validation that things are in sync via the new job in `.github/workflows/tests.yml` (this is a general pattern which could be useful for us elsewhere)

### Verification

I've done some light testing manually with the keyboard robot, though mostly I'm relying on validation from the unit tests.
It would be great to do more manual testing that the orientation data in particular is correct.

I've manually compared a small number of the multi-axis markers to the images in [april_vision's tests](https://github.com/WillB97/april_vision/tree/24b44a89065f7ac274ec624306d9d96260bb2571/tests/test_data/) and confirmed that they line up.

### Links

Fixes #377


WillB97:
You may want to look at how aplril_vision tests its marker object. It uses webots to [generate marker images with markers at known poses](https://github.com/WillB97/april_vision/blob/main/tests/webots_generator/controllers/capture_robot/capture_robot.py) and them feeds the images to a [test script to assert the calculated poses are correct](https://github.com/WillB97/april_vision/blob/main/tests/test_processor.py). Here the two halves can be combined as we're using webots' detection.

WillB97:
The order of orientation rotations are shown here: https://github.com/WillB97/april_vision/blob/main/tests/webots_generator/worlds/webots_generator.wbt#L37-L44

The maths required to convert the Webots' orientation to yaw, pitch, roll is:

```python
# Unpack the axis angle to allow for remapping and normalisation
_x, _y, _z, angle = recognition.getOrientation()
# Normalise the axis
axis_mag = hypot(_x, _y, _z)
_x, _y, _z = _x / axis_mag, _y / axis_mag, _z / axis_mag

# Remap the axis to match the kit's coordinate system
x, y, z = -_x, _y, -_z

# Calculate the intrinsic Tait-Bryan angles following the z-y'-x'' convention
# Approximately https://w.wiki/7cuk with some sign corrections,
# adapted to axis-angle and simplified
yaw = atan2(
    z * sin(angle) + x * y * (cos(angle) - 1),
    1 + (y ** 2 + z ** 2) * (cos(angle) - 1),
)
pitch = asin(x * z * (1 - cos(angle)) + y * sin(angle))
roll = atan2(
    x * sin(angle) + y * z * (cos(angle) - 1),
    1 + (x ** 2 + y ** 2) * (cos(angle) - 1),
)
```

modules/sr/robot3/vision/markers.py
PeterJCLaw, line 1: This file is adapted from what was previously `tokens.py`; it may be useful to review it as a diff against that.

modules/sr/robot3/coordinates/twin_angle.py
PeterJCLaw, line 1: This file is essentially copied from the kit API.

controllers/example_controller/keyboard_controller.py
```diff
@@ -29,7 +29,7 @@
 def angle_str(angle: float) -> str:
     if USE_DEGREES:
         degrees = math.degrees(angle)
-        return f"{degrees:.2g}°"
+        return f"{degrees:.3g}°"
```
PeterJCLaw, line 32: Not technically related, but made debugging easier.

protos/Components/SRCamera.proto
PeterJCLaw, line 1: This file hoists a common & re-usable camera which is then also used in tests.

modules/sr/robot3/vision/types.py
PeterJCLaw, line 1: Content copied from the kit API.

modules/sr/robot3/vision/convert.py
```diff
@@ -22,12 +24,7 @@ class WebotsOrientation(NamedTuple):
 
 
 def rotation_matrix_from_axis_and_angle(orientation: WebotsOrientation) -> Matrix:
```
WillB97, line 26: Are we using this anymore?

modules/sr/robot3/camera.py
```diff
@@ -80,77 +55,46 @@ def parse_marker_info(model_id: str) -> MarkerInfo | None:
     object. The digits which form the 'code' are used for determining the
     properties visible in the API.
 
-    Examples: 'F00', 'F01', ..., 'B32', 'B33', ...
+    Examples: 'F00', 'F01', ...
     """
 
-    match = MARKER_MODEL_RE.match(model_id)
+    match = MARKER_MODEL_RE.match(recognition_object.getModel())
```
WillB97, line 61: Is it actually worth having this regex instead of just setting the model to the marker number?

script/check
```diff
@@ -7,6 +7,9 @@ SCRIPTS=$(find script -type f | grep -vE "/(linting/|testing/|typing/|check$)")
 ./script/testing/test
 result=$?
 
+./script/testing/integration-test
+result=$((result | $?))
+
```
WillB97, line 12: This makes this script a lot more annoying to quickly use to check for linting and typing issues.

WillB97 (commented):
There still seems to be an awful lot of code around vision present, even though webots is doing all the pose estimation, so we should just be doing an axis conversion and calculating yaw, pitch and roll from the rotation.

modules/sr/robot3/vision/api.py
```diff
@@ -1,69 +1,82 @@
...
+    # Webots' axes nearly match ours:
+    # - x: distance away from the camera
+    # - y: distance left of the camera
+    # - z: distance *below* the camera (we want above)
+    x, y, z = recognition_object.getPosition()
+
+    marker = FiducialMarker(
+        size=recognised_object.size_m,
+        # Webots Z is inverted with regard to the one we want.
+        position=Vector((x, y, -z)),
```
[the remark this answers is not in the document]
PeterJCLaw: Good spot: 66b326210e82446983d1a56597264af74385ae3e fixes this.

modules/sr/robot3/vision/convert.py
PeterJCLaw: Yes, this is still used in setting up the `FiducialMarker` instance we return. While the orientation of that isn't used here, it will be used to validate that we're looking at the front of a marker (implementation of that will be in a separate PR). This code is validated by the new tests (for single-axis rotations) alongside the direct-to-orientation code-path.

modules/sr/robot3/camera.py
PeterJCLaw: Maybe? It provides decoupling which allows us to convey other information here or to have things which are recognisable by some cameras but not others.

script/check
PeterJCLaw: Yeah, I wasn't sure about this. I found however that I was forgetting to run the other checks while working on the integration tests so having them all together seemed preferable. It should be easy enough to use the individual scripts if people want to run the other checks.

ffc7f01e8d55f65b76ceac31a3de737f5ca006c0 speeds up the integration tests; now reliably under 3s for me (as reported by unittest, obviously webots spin-up is outside that).

I guess we could add an option/environment variable here which turns off the integration tests if people want that? I do think the default should be everything though.

Another option (that I would encourage anyway) for faster feedback is configuring your editor to run the linters in the background & show squiggles.

PeterJCLaw (commented):
Thanks for the review, I've fixed some of the issues and otherwise replied to comments.

worlds/Arena.wbt
```diff
@@ -113,7 +113,7 @@ Solid {  # Wall markers
   children [
     WallMarker {
       translation 2.154 -2.8749 0.175
-      rotation 0 0 1 0
+      rotation 0 0 1 -1.5708
       name "F0"
```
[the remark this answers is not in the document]
PeterJCLaw: https://github.com/srobo/competition-simulator/issues/419

protos/Markers/WallMarker.proto
```diff
@@ -16,7 +16,7 @@ PROTO WallMarker [
     name IS name
     model IS model
     texture_url IS texture_url
-    size 0.2 0.0001 0.2
-    visual_size 0.2 0.0001 0.2
```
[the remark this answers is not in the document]
PeterJCLaw: Note that https://github.com/srobo/competition-simulator/issues/417 already exists to remind us to do this too.

modules/sr/robot3/vision/convert.py
PeterJCLaw: > Yes, this is still used in setting up the `FiducialMarker` instance we return. While the orientation of that isn't used here, it will be used to validate that we're looking at the front of a marker (implementation of that will be in a separate PR).

That PR is https://github.com/srobo/competition-simulator/pull/420

Adimote (approved):
Looks good to me! I didn't spot anything egregious and the numbers seem to work as expected in the simulator.

PeterJCLaw:
I'm going to merge this for speed and to unblock other things. Happy to continue to discuss/address things though.

modules/sr/robot3/coordinates/twin_angle.py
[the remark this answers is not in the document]
WillB97: Twin-angle is fairly meaningless (certainly not google-able). It contains the marker's position structure so why not name it as such. Equally why has the kits marker.py been split across 3 files?

modules/sr/robot3/vision/types.py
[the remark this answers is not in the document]
WillB97: This solely contains the marker's orientation structure, why is it not named as such.

modules/sr/robot3/coordinates/twin_angle.py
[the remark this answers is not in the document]
PeterJCLaw: There are lots of different ways that the position of an entity could be spelled and has been spelled over time. The code here accounts for all of those spellings (not the existence of polar and cartesian coordinate types in sibling files). "Position" is a very vague term for this. I can see why the name was chose in the kit (and thus this codebase is stuck with it), however I think it's very uninformative (and just as non-googleable as "twin angle"). Twin-angle at least describes the kind of coordinates we're looking at.

> Equally why has the kits marker.py been split across 3 files

This could equally be asked the other way around, but I'm not sure it's useful to do so. The kit and simulator APIs are different codebases. There is some benefit to them sharing code, but that is not an end in itself. This codebase has now outlasted three[^1] generations of the kit codebase, so I don't think it's reasonable to expect it to chop & change as the kit API is rebuilt.

The location of the `Position` type is within the portion of the API which is independent from Webots and is thus able to be tested in isolation. Until very recently (i.e: the latter part of this PR) this was the only way to have unit tests for the API code. It remains the more efficient and more easily debuggable way to have tests.

[^1]: ish, depends a bit how you count

modules/sr/robot3/vision/types.py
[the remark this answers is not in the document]
PeterJCLaw: Because it does not contain any of the logic relating to the orientation -- only the type. (The logic is in `convert.py`; itself not the best name, but I'm avoiding making this PR even bigger by renaming that too)

PR #416, opened 2023-09-26
An alternate Arduino implementation

### LINKED ISSUE #380: Ruggeduino should not error about unused pins
It should error about pins out of bounds of a real ruggeduino, but not ones in between which happen to be unused.

### LINKED ISSUE #411: Update Arduino API to match SR2024 kit
The Arduino API has changed for SR2024 and thus needs updating.

WillB97:
An alternate implementation to #412 based more closely on the [sr-robot API](https://github.com/srobo/sr-robot/blob/a1c7d2e5f21d1d482d2b2ff1d99110865ecb392e/sr/robot3/arduino.py). Device wrappers are used in place of the serial wrapper to interface the pin to the webots device.

Fixes #411
Fixes #380

The first commit renames the ruggeduino to arduino.

Tested by manually using the keyboard robot.

### DIFF
--- a/modules/sr/robot3/arduino.py
+++ b/modules/sr/robot3/arduino.py
@@ -10,6 +10,7 @@ from controller import Robot
 from sr.robot3.utils import BoardIdentity
 from sr.robot3.arduino_devices import (
     Led,
+    EmptyPin,
     PinDevice,
     Microswitch,
     DistanceSensor,
@@ -194,8 +195,9 @@ class Pin:
         self._disabled = disabled
         self._mode = GPIOPinMode.INPUT
 
+        # The device is set to an empty pin by default.
         if device is None:
-            device = PinDevice.empty()
+            device: PinDevice = EmptyPin()
         self._device = device
 
     @property
@@ -288,5 +290,5 @@ class Pin:
         return (
             f"<{self.__class__.__qualname__} "
             f"index={self._index} analog={self._supports_analog} "
-            f"disabled={self._disabled} {self._device.__class__.__name__}>"
+            f"disabled={self._disabled} {self._device!r}>"
         )
--- a/modules/sr/robot3/arduino.py
+++ b/modules/sr/robot3/arduino.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 import logging
 from enum import Enum, IntEnum
 from types import MappingProxyType
-from typing import Optional
 
 from controller import Robot
 from sr.robot3.utils import BoardIdentity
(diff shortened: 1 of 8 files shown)


modules/sr/robot3/arduino.py
```diff
@@ -0,0 +1,289 @@
...
+        self._pins[AnalogPins.A3]._connect_device(DistanceSensor(webot, 'Right DS'))
+        self._pins[AnalogPins.A4]._connect_device(DistanceSensor(webot, 'Front DS'))
+        self._pins[AnalogPins.A5]._connect_device(DistanceSensor(webot, 'Back DS'))
+        # NOTE: these pins don't exist on a real arduino.
+        self._pins[AnalogPins.A6]._connect_device(
+            PressureSensor(webot, 'finger pressure left'),
+        )
+        self._pins[AnalogPins.A7]._connect_device(
+            PressureSensor(webot, 'finger pressure right'),
+        )
```
PeterJCLaw, line 110: Configuration of which pins are doing what is not an inherent part of the Arduino and should probably be passed in as an argument.

modules/sr/robot3/arduino.py
```diff
@@ -0,0 +1,289 @@
...
+
+        :return: The identity of the board.
+        """
+
+        return BoardIdentity(
+            manufacturer="Student Robotics",
+            board_type="SRduino",
+            asset_tag=self._serial_num,
+            sw_version='1.0',
+        )
```
PeterJCLaw, line 126: This is redundant in the simulator. (It's also kinda wrong)
```suggestion
```

modules/sr/robot3/arduino.py
```diff
@@ -0,0 +1,289 @@
...
+    __slots__ = ('_serial_num', '_serial', '_pins', '_identity')
+
+    @staticmethod
+    def get_board_type() -> str:
+        """
+        Return the type of the board.
+
+        :return: The literal string 'Arduino'.
+        """
+        return 'Arduino'
```
PeterJCLaw, line 73: This is redundant in the simulator. (It's also kinda wrong)
```suggestion
```

modules/sr/robot3/utils.py
```diff
@@ -1,13 +1,32 @@
...
+    :param asset_tag: The asset tag of the board,
+        this should match what is printed on the board
+    :param sw_version: The firmware version of the board
+    """
+    manufacturer: str = ""
+    board_type: str = ""
+    asset_tag: str = ""
+    sw_version: str = ""
+
+
```
PeterJCLaw, line 29: This is redundant in the simulator.
```suggestion
```

modules/sr/robot3/arduino_devices.py
```diff
@@ -0,0 +1,152 @@
...
+)
+from sr.robot3.utils import map_to_range, get_robot_device
+from controller.device import Device
+from sr.robot3.randomizer import add_jitter
+from sr.robot3.output_frequency_limiter import OutputFrequencyLimiter
+
+LOGGER = logging.getLogger(__name__)
+
+
+class PinDevice(abc.ABC):
```
PeterJCLaw, line 21: If we're going to use this approach (devices separate from pins), which I admit is probably a better simulation of the real robot, then I think we should model the simulation as such. That is: allow for any device being used on any pin and have them communicate in a way that reflects reality.

Specifically I think we should have _only_ analogue-read and digital-write operations at the device level. We can do the analogue-to-digital conversion consistently in the `Pin` type, with a suitable threshold.

This approach avoids all the quandaries of whether a given type should implement a given interface or whether returning analogue zero from a microswitch is correct. If we're modelling these as separate devices then modelling them as _real_ analogue devices is the way to go.

modules/sr/robot3/arduino_devices.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    A device connected to a pin on the Arduino.
+    """
+    _ANALOG_MIN = 0.0  # Volts
+    _ANALOG_MAX = 5.0
+    _DEVICE_TYPE: Optional[type[Device]] = None
+    _webot_device: Optional[Device]
```
PeterJCLaw, line 28: style suggestion -- it would be nicer to separate out the webots nature of the devices from their device simulation nature. This would more easily allow for non-webots devices, such as the null device (`EmptyPin`).

One approach might be a base type which just implements the base API and has a name, another might be to use a protocol and decouple the inheritance chains of the non-webots devices.

modules/sr/robot3/arduino.py
```diff
@@ -0,0 +1,289 @@
...
+from sr.robot3.arduino_devices import (
+    Led,
+    EmptyPin,
+    PinDevice,
+    Microswitch,
+    DistanceSensor,
+    PressureSensor,
+)
+
+logger = logging.getLogger(__name__)
```
PeterJCLaw, line 19: This appears to be unused.

modules/sr/robot3/arduino.py
```diff
@@ -0,0 +1,289 @@
...
+
+        :param pin_number: The pin number to encode.
+        :return: The pin number in the serial format.
+        :raises ValueError: If the pin number is invalid.
+        """
+        try:  # bounds check
+            self.pins[pin_number]._check_if_disabled()
+        except (IndexError, IOError):
+            raise ValueError("Invalid pin provided") from None
+        return chr(pin_number + ord('a'))
```
PeterJCLaw, line 160: These appear to be unused (and aren't part of the documented API)
```suggestion
```

modules/sr/robot3/arduino.py
WillB97: 813d119db0f894315bfc2ace5ab8e45c69a0dc9d

modules/sr/robot3/arduino.py
WillB97: 813d119db0f894315bfc2ace5ab8e45c69a0dc9d

modules/sr/robot3/arduino.py
WillB97: This function is present in the kit API and returns a suitable value here to be compatible. This a step towards students being able to use almost the same code in the simulator as on the physical robot without it crashing due to undefined functions.

modules/sr/robot3/arduino.py
WillB97: As above, this function is present in the kit API.

modules/sr/robot3/utils.py
WillB97: This is used to support the above function.

modules/sr/robot3/arduino.py
WillB97: As above, this function is present in the kit API. `command` is part of the extended SR firmware.

modules/sr/robot3/arduino_devices.py
WillB97: What we've got here works for the moment, we can always re-evaluate if we find a need for additional non-webots devices.

modules/sr/robot3/arduino_devices.py
WillB97: The purpose of the simulation is to model the real world in a simplified way such that the behaviour of a system can be understood. The simplification here is to make is clear how a device should be used; a distance sensor is analog, a micro-switch is digital.

A warning is now printed when one of the default methods are used to warn competitors that this is not how the device is meant to be used.

modules/sr/robot3/arduino.py
PeterJCLaw: I appreciate the desire to enable similar code between the two environments, however we should also bear in mind that any code which we have comes with a cost. This is especially true for code which is untested (as the majority of the code in the simulator unfortunately is). I would greatly prefer that we not introduce code which is not actually necessary to the operation of the simulator nor documented as part of the kit API.

Aiming for the entire code surface between the two environments being the same seems to me like it's going to create far more opportunities for things to get out of step than it will aid.

Unless the plan is also to update the kit API to include all the utility functions which the simulator needs/has/provides?

modules/sr/robot3/arduino_devices.py
PeterJCLaw: What we have in #411 _works_, if that's the bar we're aiming for?
I had thought the concerns with that approach were around not sufficiently well separating out the various modelling concerns. Perhaps I'd misunderstood?

modules/sr/robot3/arduino.py
8BitJosh: One of the goals of this project was to bring into line the physical kit and simulator interfaces. To do this we should be including all the public interfaces accessable to the user. 
These functions are very basic and stub out the functions. They may be unnessasery but it will stop the students code from crashing if they run code between the robot and the simulator, I believe this to be valuable addition for a UX perspective. 

I think this approach of taking the kit API code and swapping the serial device out for a webots device is a good method. If the structure remains similar between the two, it will ease the development process.

8BitJosh: I also think this:

> It's also kinda wrong

Comes off as condesending, please in the future could it be replaced with a clear explaination of what you mean. I for one do not understand what you are refering to.

modules/sr/robot3/arduino_devices.py
8BitJosh: Yes both of the PRs work, however I think this implimentation is clearer than #411 
I think Will was refering to the fact at the moment we do not have any non-webots devices apart from the EmptyPin. If this were to change we can then revisit this discussion to what best suits the new device we want to add

modules/sr/robot3/arduino.py
PeterJCLaw: > One of the goals of this project was to bring into line the physical kit and simulator interfaces.

I disagree with the depth that this is being taken to. Perhaps I'm mistaken, but it doesn't match understanding of the expectations we currently have for the simulator. I'd be happy to chat about this, however I would rather that this not become a distraction from getting the simulator to a point where it's usable for SR2024.

> > It's also kinda wrong
> 
> Also comes off as condesending, please in the future could it be replaced with a clear explaination of what you mean. I for one do not understand what you are refering to.

Fair.

> a clear explaination of what you mean.

The properties presented in the snippet I've suggested be removed don't make sense in the simulator. There is no board, so there is no manufacturer, no software to have a version of nor asset tag which goes with it. If we are going to present mocks of these values in the simulator (which is perhaps reasonable if we really want the API here to be exhaustive, though I don't think that's a good idea) then we should prefer to make them actually distinct from the real kits -- this is the approach taken by the existing metadata APIs where they exist (though mostly they don't).

PeterJCLaw:
Changes similar to those proposed here were worked into #412, thanks!

PR #434, opened 2023-10-14
Disable physics on markers so things can't get stuck to tokens

PeterJCLaw:
This prevents an issue where parts of the robot grabber could get stuck on the side of tokens, we believe due to the grabber getting "between" the marker and the token physics objects. Disabling the physics on the marker means we can't have a bounding box there, which in turn means the visible bounds of the recognition objects are quite a bit wider than they ought to be.

Happily that doesn't affect the parts of the vision data that we present to competitors. It does affect our occlusion logic though, meaning that we need to disable that. From what I can recall our inclusion of that was a workaround for recognition objects being visible through other recognition objects, which is not an issue now that we have distinct entities for the recognition objects (i.e: markers) and the things they're attached to (tokens, props, etc.).

Builds on #433.

### DIFF
--- a/modules/sr/robot3/vision/api.py
+++ b/modules/sr/robot3/vision/api.py
@@ -70,15 +70,9 @@ def markers_from_objects(
         key=lambda x: x[0].position.magnitude(),
     )
 
-    preceding_rectangles: list[Rectangle] = []
     markers = []
-    for marker, image_rectangle, recognised_object in markers_with_info:
-        if (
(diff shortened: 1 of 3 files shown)


Adimote (approved):
LGTM!

Adimote:
I verified that this works, I cannot reproduce the issue of markers getting stuck, and it looks like the marker IDs do still follow occlusion rules

PR #435, opened 2023-10-16
Minimally prepare the arena for a point-zero release

PeterJCLaw:
This is stripped out enough to allow us to do a release which will let teams confirm that they have installed the simulator correctly, without containing anything specific to this year's game.

Review in part also for "is this a good idea".

### DIFF
--- a/worlds/Arena.wbt
+++ b/worlds/Arena.wbt
@@ -2,11 +2,7 @@
 
 EXTERNPROTO "../protos/Markers/WallMarker.proto"
 EXTERNPROTO "../protos/Robot_2022/Robot_2022.proto"
-EXTERNPROTO "../protos/Tokens/SRToken_Gold.proto"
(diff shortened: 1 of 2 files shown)


WillB97:
Doesn't this require teams to install the simulator twice in a period of 5 days?

PeterJCLaw:
> Doesn't this require teams to install the simulator twice in a period of 5 days?

Why? We're not referencing a different version of Webots.

PeterJCLaw:
If you mean the simulation (i.e: the world etc.) then yes, however the simulation files are small and upgrading them is something that the teams should be able to do anyway. (If they can't then it's probably better that they find that out at Kickstart rather than later).

Adimote (approved):

PeterJCLaw:
We ended up not shipping this.

PR #439, opened 2023-10-17
Support access to unused servo slots

PeterJCLaw:


Adimote (approved):
This change makes a lot of sense!

PR #440, opened 2023-10-17
Provide a hook to use the keyboard robot in other zones

PeterJCLaw:
This makes it slightly more obvious how this robot can be used in other zones, while providing a clear message about why it won't work by default.

Fixes https://github.com/srobo/competition-simulator-private/issues/61

Builds on #437 to avoid conflicts.

### DIFF
--- a/controllers/example_controller/keyboard_controller.py
+++ b/controllers/example_controller/keyboard_controller.py
@@ -26,8 +26,8 @@ CONTROLS = {
 # place to remap other zones if desired. Modify the keys of the dictionary to
 # the zones you want to use.
 ZONE_MAP = {
-    1: 0,
-    0: 1,
+    0: 0,
+    1: 1,
 }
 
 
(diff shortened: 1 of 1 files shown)


controllers/example_controller/keyboard_controller.py
```diff
@@ -22,6 +22,18 @@
     "angle_unit": (ord("B"), ord("B")),
 }
 
+# Map robot zone to control set. Defaults to straight mapping, but provides a
+# place to remap other zones if desired. Modify the keys of the dictionary to
+# the zones you want to use.
+ZONE_MAP = {
```
Adimote, line 28: Shouldn't this be:
```
0: 0,
1: 1,
```

So zone 0 has WSAD controls and zone 1 has IJKL controls?

Otherwise it'll change the behavior of the robot

Adimote (approved):

controllers/example_controller/keyboard_controller.py
PeterJCLaw: Gah, oops. Yeah, it was like that and then I was testing it and apparently committed the test code.

9988927

PR #441, opened 2023-10-17
Improve self-limiting of keyboard robot inputs

PeterJCLaw:
I've occasionally seen errors from non-boosted inputs which appear to be too large for the motors to handle. The code seems to allow that to happen if there's more than one input character in the buffer. Clamping all inputs (not just boosted ones) should make these errors impossible.


Adimote (approved):
I didn't even notice there was a boost function!

PR #447, opened 2023-10-24
Remove unused robots in dev mode too

Adimote:
I'm not entirely sure what the side-effects are of reordering these events, advice would be more than welcome!

But the benefits of this is my simulator runs at 5x speed with 1 robot in the arena, and 3x speed with 4 robots in the arena, so it's about a 66% simulation speed improvement

### DIFF
--- a/controllers/competition_supervisor/competition_supervisor.py
+++ b/controllers/competition_supervisor/competition_supervisor.py
@@ -209,8 +209,6 @@ def run_match(supervisor: Supervisor) -> None:
 
 
 def main() -> None:
-    quit_if_development_mode()
-
     controller_utils.tee_streams(
         controller_utils.get_competition_supervisor_log_filepath(),
     )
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
Immediate thoughts, but haven't thought much yet:
- we should test this in each of the cases of
 - the default robot code location
 - just `zone-0`
 - just some other zone
 - what happens if there is no user-code found
- does this make it less obvious to competitors that there will be other robots in the arena?
- does this make it less obvious to competitors that they can test their code in other starting zones?

Adimote:
>   * the default robot code location
Robot was deleted :( will have to fix before submitting.

>   * just `zone-0`
This works

>   * just some other zone
This also works

>   * what happens if there is no user-code found

Seems to behave correctly, a robot.py was created!


> * does this make it less obvious to competitors that there will be other robots in the arena?

I don't think so? Maybe we can mitigate this by adding a starting box under where the robots start

> * does this make it less obvious to competitors that they can test their code in other starting zones?

I think we should fix this by documenting it better

PR #450, opened 2024-01-21
Improve windows support around run-comp-match

PeterJCLaw:
I'm pretty confident we want the first change here (docs)~, less sure about the second. The latter definitely needs testing by someone on Windows. I can probably do that at some point, but I don't know if I have webots setup/etc. on my Windows machine so if someone who does is able to review that before I get back to this please do so!~

Have now tested this on Windows & fixed a small issue. In my testing I didn't have the right Python packages installed so the simulation only got as far as that check, however I think that's still far enough to prove that the rest of the system worked -- it was able to execute Python to run the match script, which found Webots via `WEBOTS_EXECUTABLE` which was subsequently able to launch Python to run the "user" code.

Testing was based on a manually taken copy of the batch script dropped into an existing copy I had of the sr2024.1 release.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,7 +54,7 @@ jobs:
       - name: Define scripts to test
         id: define_scripts
         run: |
-          SCRIPTS=$(find script -type f | grep -vE "/(linting/|testing/|typing/|check$)" | tr "\n" " ")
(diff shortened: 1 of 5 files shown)


prophile (approved):

PeterJCLaw:
@Adimote confirmed this works too :)

PR #451, opened 2024-01-22
Fix lints

PeterJCLaw:


controllers/lighting_controller/lighting_controller.py
```diff
@@ -224,6 +224,7 @@ def do_lighting_step(self) -> None:
                 self.luminosity_fade.current_luminosity,
             )
 
+        to_remove: list[LightFade]
         for fade in self.lighting_fades:
```
WillB97, line 228: Why aren't we just doing `self.lighting_fades.copy()`, as is done elsewhere in this file?

PeterJCLaw: It didn't occur to me to take that approach. (This approach tends to be slightly more efficient as it avoids a full copy of the list)

---
Is the order of these operations significant? If not, then `for fade in reversed(self.lighting_fades)` would be even more efficient while still avoiding the iteration issues.

Adimote (approved):

